PR #15, opened 2016-05-13
AyahInfo SQLite methods

ahmedre:
This is a pull request which starts implementing some of the methods for reading AyahInfo from the sqlite database. My intention for this is to get some feedback before finishing this up and making it commit-worthy. Please note that I have never written Swift before, so please feel free to be as critical as possible, and insha'Allah after a few rounds, I should get the hang of it. I'll comment below on my reasons for doing things the way I did (and ask questions about whatever I want to ask about insha'Allah).

As an aside, this PR actually has 2 commits - the first just imports the pod (large changed file count mainly comes from that commit), and the second just does some very very basic AyahInfo work in progress stuff.


Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,38 @@
...
+struct AyahInfoPersistenceStorage {
+
+    private let db: Connection
+    private let glyphsTable = Table("glyphs")
+
+    private let idColumn = Expression<Int>("glyph_id")
+    private let suraColumn = Expression<Int>("sura_number")
+    private let ayahColumn = Expression<Int>("ayah_number")
+
+    private let imageSize = "1920"
```
ahmedre: where should this live since it's shared by various parts of the app?

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    init() throws {
+        let file = String(format: "images_\(imageSize)/databases/ayahinfo_\(imageSize)")
+        let path = NSBundle.mainBundle().pathForResource(file, ofType: "db")!
```
ahmedre: what's the correct way to do this without forcing unwrapping?

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Created by Ahmed El-Helw on 5/12/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+struct AyahInfoPersistenceStorage {
+
+    private let db: Connection
```
ahmedre: should this be explicitly static, or is it ok to just have one, since we should only ever instantiate one (which will be part of the corresponding view controller?)

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Quran
+//
+//  Created by Ahmed El-Helw on 5/12/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+struct AyahInfoPersistenceStorage {
```
ahmedre: i understand that your intention was for this to adopt the `SQLitePersistenceStorage` protocol. however, some thoughts - it seems to me that the library i pulled for SQLite attempts to be more expressive, and while it gives the person the option to write sql, it doesn't require it (as you can see in the example below).

that having been said, i am wondering if you'd be okay with `AyahInfoPersistenceStorage` which just has methods like `getAyahInfoForAyah`, `getAyahInfoForPage`, and `getAyahInfoForXY` for example instead of actually having to map this to something used by `execute`.

moreover, 2 other questions:
1. who is responsible for calling this on a background thread? i saw you did some stuff with this for the various `DataRetriever` classes for example, but wondering about this.
2. if these methods are very low level (i.e. given some parameters, return a list of `AyahInfo` objects, for example, then where does "business logic" on top of these methods go (for example, an ayah may have multiple `AyahInfo` entries (one per line), so what if we want to combine them into one `AyahInfo`, where would that go?

ahmedre:
btw, i tested this by adding:

```swift
        let x = try? AyahInfoPersistenceStorage()
        try? x?.getAyahInfo(1, ayah: 1) {_ in
        }
```

within one of the view controllers.

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    private let db: Connection
+    private let glyphsTable = Table("glyphs")
+
+    private let idColumn = Expression<Int>("glyph_id")
```
ahmedre: should these be within a nested struct?

mohamede1945:
@ahmedre Thank you so much for starting this PR. That would be so helpful for me. I'll start answering individual questions now.

Quran/AyahInfoPersistenceStorage.swift
mohamede1945: In general, I'm against singleton. So, I suppose the latter is the way to go since we will only need one that will be owned indirectly by a view controller.

Quran/AyahInfoPersistenceStorage.swift
mohamede1945: My first intention was having an abstraction layer to abstract the SQLite library or let's say, driver. So that we can replace it later with minimal changes. Then, we will have an Ayah Info class that will use the abstraction layer. It will be the same for Gapless Qari sqlite database.

So, `SQLitePersistenceStorage` should be a protocol that could have one implementation with sqlite.swift then, in the future if we find another library we could use it then all we need to do is just implement SQLitePersistenceStorage and update the dependency injection container to use the new one. Feel free to modify SQLitePersistenceStorage protocol methods signatures that best suit our needs. I think it could be abstracted better like having a similar API as Sqlite.swift has.

Then we could have AyahInfoRetriever protocol that does what you said like `getAyahInfoForAyah`, `getAyahInfoForPage`, and `getAyahInfoForXY`. Then the one implementation called `SQLiteAyahInfoRetriever` that uses `SQLitePersistenceStorage`.

Then we will have another one for Qari database like `QariTimingRetriever` and `SqliteQariTimingRetriever`. 

But I'm also totally fine with your approach.

mohamede1945:
Regarding the BTWs comment :)

1. Great that you were able to test it.
2. Yeah, that's totally fine. 
3. Yes, it has pros and cons. 
From https://guides.cocoapods.org/using/using-cocoapods.html we can read a summary of the pros and cons, but the cocoapods team itself recommend adding it.

> We recommend that you keep the Pods directory under source control, and don't add it to your .gitignore. 

The 2 main reasons I've for keeping pods under source control are:
1. Two collaborators don't end up with different versions of different pods. Then one see a bug in the code the other don't see it. Because one did pod update for example.
2. Pods unavailability, for example a person decided to completely delete the repo of a pod.

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,43 @@
...
+    private let pageColumn = Expression<Int>("page_number")
+    private let suraColumn = Expression<Int>("sura_number")
+    private let ayahColumn = Expression<Int>("ayah_number")
+
+    private let imageSize = "1920"
+
+    init() throws {
+        let file = String(format: "images_\(imageSize)/databases/ayahinfo_\(imageSize)")
+        guard let path = NSBundle.mainBundle().pathForResource(file, ofType: "db") else { fatalError() }
+        db = try Connection(path, readonly: true)
```
mohamede1945: I would like to wrap that exception into a PersistenceError. Also, what about lazily initialize the connection.

mohamede1945:
It seems like you already know swift pretty well. Really.

mmahalwy:
wowzah! @ahmedre woot!

ahmedre:
closing and re-opening a new one since i messed up the history :(

PR #16, opened 2016-05-14
AyahInfo Database Handling

ahmedre:
These patches add support for reading the ayahinfo database to figure out which verse was long pressed.

This patch has 2 parts:
1. import SQLite.swift
2. sqlite related code (updated based on discussion from #15).


Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+//  Created by Ahmed El-Helw on 5/12/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+struct AyahInfoPersistenceStorage {
+
+    struct Columns {
```
mohamede1945: Can we make the struct private?

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    private let db: Connection
+    private let imageSize = "1920"
+    private let glyphsTable = Table("glyphs")
+    private let columns = Columns()
+
+    init() throws {
+        let file = String(format: "images_\(imageSize)/databases/ayahinfo_\(imageSize)")
+        guard let path = NSBundle.mainBundle().pathForResource(file, ofType: "db") else { fatalError() }
+        db = try Connection(path, readonly: true)
```
mohamede1945: Can we lazily initialize the connection?

mohamede1945: Also, I would love to wrap the error comming from creating the connection into `PersistenceError`. So, we can show appropriate error message to the user in case of failure.

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    private let db: Connection
+    private let imageSize = "1920"
+    private let glyphsTable = Table("glyphs")
+    private let columns = Columns()
+
+    init() throws {
+        let file = String(format: "images_\(imageSize)/databases/ayahinfo_\(imageSize)")
+        guard let path = NSBundle.mainBundle().pathForResource(file, ofType: "db") else { fatalError() }
```
mohamede1945: Can you add a message to the `fatalError`?

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+//  Quran
+//
+//  Created by Ahmed El-Helw on 5/12/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import Foundation
+import SQLite
+
+struct AyahInfoPersistenceStorage {
```
mohamede1945: Can you create a protocol for `AyahInfoPersistenceStorage` then making that struct conforms to it with a name like `SQLiteAyahInfoPersistenceStorage`? Then we can dependency inject it to the AyahRetriever instea of depending on implementation.

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+        let file = String(format: "images_\(imageSize)/databases/ayahinfo_\(imageSize)")
+        guard let path = NSBundle.mainBundle().pathForResource(file, ofType: "db") else { fatalError() }
+        db = try Connection(path, readonly: true)
+    }
+
+    func getAyahInfoForPage(page: Int) throws -> [AyahNumber : [AyahInfo]] {
+        let query = glyphsTable.filter(columns.page == page)
+
+        var result = [AyahNumber : [AyahInfo]]()
+        for row in try db.prepare(query) {
```
mohamede1945: Can we wrap the error into PersistenceError? So, we can show an appropriate error message to the user in case of failure.

Quran/AyahInfoPersistenceStorage.swift
```diff
@@ -0,0 +1,70 @@
...
+        }
+        return result
+    }
+
+    func getAyahInfoForSuraAyah(sura: Int, ayah: Int) throws -> [AyahInfo] {
+        let query = glyphsTable.filter(columns.sura == sura && columns.ayah == ayah)
+
+        var result: [AyahInfo] = []
+        let ayah = AyahNumber(sura: sura, ayah: ayah)
+        for row in try db.prepare(query) {
```
mohamede1945: Can we wrap the error into PersistenceError? So, we can show an appropriate error message to the user in case of failure.

Quran/QuranViewController.swift
```diff
@@ -193,6 +204,45 @@ class QuranViewController: UIViewController, AudioBannerViewPresenterDelegate {
         setBarsHidden(navigationController?.navigationBarHidden == false)
     }
 
+    func onViewLongTapped(sender: UILongPressGestureRecognizer) {
+        guard sender.state == UIGestureRecognizerState.Ended else { return }
```
mohamede1945: 1. This code assumes that the image is full screen which is not the case. Actually, the only class responsible for laying out the image is `QuranPageCollectionViewCell` and accordingly these calculations should be there and using `mainImageView.bounds.size` and `mainImageView.image.size` with no hard coded values.

2. Loading an image is async. Sometimes the screen becomes blank and we shouldn't handle long press there. So, I think the gesture recognizer needs to move as well to `QuranPageCollectionViewCell` with a delegate handling `gestureRecognizerShouldBegin` with `mainImageView.image.image != nil`.

Quran/SQLiteAyahInfoRetriever.swift
```diff
@@ -10,9 +10,14 @@ import Foundation
 
 struct SQLiteAyahInfoRetriever: AyahInfoRetriever {
 
-    let persistence: SQLitePersistenceStorage
+    let persistence: AyahInfoPersistenceStorage
 
     func retrieveAyahsAtPage(page: Int, onCompletion: Result<[AyahNumber : [AyahInfo]], PersistenceError> -> Void) {
-        unimplemented()
+        Queue.background.async {
+            let result = try? self.persistence.getAyahInfoForPage(page)
```
mohamede1945: We need to handle both success and failure cases.

mohamede1945:
Thanks so much for your efforts.

PR #17, opened 2016-05-16
Support highlighting a particular ayah

ahmedre:
This patch consists of two parts - first, normalizing the results that
come back from the database so that they appear consistent per line
(since the database stores positions per word). Second, it implements
the HighlightingView so that verses can be highlighted.

### DIFF
--- a/Quran/AyahInfo.swift
+++ b/Quran/AyahInfo.swift
@@ -9,16 +9,29 @@
 import Foundation
 
 struct AyahInfo {
-    let pageNumber: Int
+    let page: Int
+    let line: Int
     let ayah: AyahNumber
     let position: Int
     let minX: Int
     let maxX: Int
     let minY: Int
     let maxY: Int
+
     var rect: Rect {
         get {
             return Rect(x: minX, y: minY, width: maxX - minX, height: maxY - minY)
         }
+    }
+
+    func engulf(other: AyahInfo) -> AyahInfo {
+        return AyahInfo(page: page,
+                        line: line,
+                        ayah: ayah,
+                        position: position,
+                        minX: min(self.minX, other.minX),
+                        maxX: max(self.maxX, other.maxX),
+                        minY: min(self.minY, other.minY),
+                        maxY: max(self.maxY, other.maxY))
     }
 }
--- a/Quran/HighlightingView.swift
+++ b/Quran/HighlightingView.swift
@@ -12,7 +12,51 @@ import UIKit
 // It's also expected to reuse layers instead of dropping & creating new ones.
 class HighlightingView: UIView {
 
-    @IBInspectable var highlightColor: UIColor = UIColor.redColor()
+    @IBInspectable var highlightColor: UIColor = UIColor(red: 0.275, green: 0.58, blue: 0.651, alpha: 0.25)
 
-    var highlightableAreas: [Rect] = []
+    var highlightedAyat: Set<AyahNumber> = Set<AyahNumber>() {
+        didSet {
+            setNeedsDisplay()
+        }
+    }
+
+    var ayahInfoData: [AyahNumber: [AyahInfo]]? {
+        didSet {
+            setNeedsDisplay()
+        }
+    }
+
+    private var imageScale: CGFloat = 0.0
+    private var xOffset: CGFloat = 0.0
+    private var yOffset: CGFloat = 0.0
+
+    override func drawRect(rect: CGRect) {
+        super.drawRect(rect)
+        guard highlightedAyat.count > 0 else { return }
+
+        let context = UIGraphicsGetCurrentContext()
+        for ayah in highlightedAyat {
+            guard let ayahInfo = ayahInfoData?[ayah] else { continue }
+
+            for piece in ayahInfo {
+                CGContextSetFillColorWithColor(context, highlightColor.CGColor)
+                CGContextFillRect(context, piece.rect.applyScale(imageScale, xOffset: xOffset, yOffset: yOffset))
+            }
+        }
+    }
+
+    func setScaleInfo(scale: CGFloat, xOffset: CGFloat, yOffset: CGFloat) {
+        self.imageScale = scale
+        self.xOffset = xOffset
+        self.yOffset = yOffset
+        setNeedsDisplay()
+    }
+
+    func reset() {
+        highlightedAyat = Set<AyahNumber>()
+        ayahInfoData = nil
+        imageScale = 0.0
+        xOffset = 0.0
+        yOffset = 0.0
+    }
 }
--- a/Quran/QuranPageCollectionViewCell.swift
+++ b/Quran/QuranPageCollectionViewCell.swift
@@ -23,7 +23,6 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
     @IBOutlet weak var scrollView: UIScrollView!
 
     var page: QuranPage?
-
     var sizeConstraints: [NSLayoutConstraint] = []
 
     override func awakeFromNib() {
@@ -34,6 +33,7 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
     override func prepareForReuse() {
         super.prepareForReuse()
         scrollView.contentOffset = CGPoint.zero
+        highlightingView.reset()
     }
 
     override func layoutSubviews() {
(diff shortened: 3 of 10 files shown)


ahmedre:
the easiest way to test this is to add:
`cell.highlightAyah(AyahNumber(sura: 1, ayah: 2))`
after `cell.setAyahInfo` in `QuranPagesDataSource.swift`.

one issue that i noticed is that sometimes, the scale is not correct (especially when switching between portrait and landscape, and/or sometimes on the first time the view is entered).

screenshots:

simulator screen shot may 15 2016 5 51 46 pm

simulator screen shot may 15 2016 5 56 08 pm

mohamede1945:
That's an amazing work! Thank you so much for that! 

I will have a look, but I assume the issue is that the scale is set not immediately, but when the UIKit tries to layout the views (i.e. in `LayoutSubviews`) accordingly, I would expect that we re-draw the highlightable areas there. I'll have a look at the code now.

Quran/HighlightingView.swift
```diff
@@ -12,7 +12,36 @@ import UIKit
 // It's also expected to reuse layers instead of dropping & creating new ones.
```
mohamede1945: For this class, I would expect a re-draw if any of the properties change. Something like that
```
var highlightedAyat: [AyahNumber] = [] {
  didSet {
    setNeedsDisplay()
  }
}
```
The same for `ayahInfoData`,  `imageScale` and `yOffset`.

Quran/HighlightingView.swift
```diff
@@ -12,7 +12,36 @@ import UIKit
...
+    }
+
+    func setScale(scale: CGFloat, yOffset: CGFloat) {
+        self.imageScale = scale
+        self.yOffset = yOffset
+        setNeedsDisplay()
+    }
+
+    func reset() {
+        highlightedAyat = []
```
mohamede1945: Can we also reset `imageScale` and `yOffset` to 0 for consistency reasons?

mohamede1945:
I'm trying to figure the issue now. It always occurs on Portrait and works fine on Landscape.

mohamede1945:
I tried but wasn't able to figure out the problem.

ahmedre:
i found the problem and a meh solution, but thought to run it by you since i think you can think of a cleaner way to do it.

the problem is basically that the subviews have not yet been laid out yet, and this affects the calculation (and, unfortunately, the cell doesn't have a `viewDidLayoutSubviews`). so my fix propagates the event for `viewDidLayoutSubviews` to the data source, which then forces a re-computation of the scale.

if you run my patch and jump directly to the page where there's a highlight, you'll see the highlight wrong for a second, and then it jumps to the correct place (this can be fixed by "don't highlight until we've gotten `viewDidLayoutSubviews`, but i'll wait for your suggestion on a good way to fix this insha'Allah).

ahmedre:
(intentionally didn't rebase for simplicity - will do so after hearing your suggestions insha'Allah).

mohamede1945:
I just downloaded the code and I still see the same problem. Here is a video https://dl.dropboxusercontent.com/u/60435436/highlight.mov

What I'm highlighting after retrieving the info:
```
cell.setAyahInfo(data.value)
cell.highlightAyah(AyahNumber(sura: 1, ayah: 2))
```

So, you might have it different or something. Please let me know.

Regarding `viewDidLayoutSubviews`, its counterpart in a UIView subclass is `layoutSubviews`. In my opinion, I think it's a problem with the calculations because it works on landscape perfectly well.

ahmedre:
your video is showing what i would expect, i think you missed my comment:

>>> if you run my patch and jump directly to the page where there's a highlight, you'll see the highlight wrong for a second, and then it jumps to the correct place (this can be fixed by "don't highlight until we've gotten viewDidLayoutSubviews, but i'll wait for your suggestion on a good way to fix this insha'Allah).

ahmedre:
update and rebase per offline discussion - everything is working now al7amdulillah.

simulator screen shot may 17 2016 10 48 11 pm

mohamede1945:
It's working great now! Thank you so much for that. 

Now, the way to highlight ayahs will be to save a set of ayahs in the data source. Then in configuring the cell, I will set these values to the cell for example `cell.highlightAyah(AyahNumber(sura: 1, ayah: 1))` after `cell.mainImageView.image = nil`. But it is not working. I suggest to do the following:

1. In `QuranPageCollectionViewCell`inside `func highlightAyah(ayah: AyahNumber)`, we don't need the checks for `ayahInfo` not nil and `if ayahInfo.count > 0`.

I will mention them inline in the code.

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -23,7 +23,7 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
     @IBOutlet weak var scrollView: UIScrollView!
 
     var page: QuranPage?
-
+    var ayahInfoData: [AyahNumber: [AyahInfo]]?
```
mohamede1945: We don't need `ayahInfoData`, since it is already in the `HighlightingView` view and we don't need it here at all.

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -50,5 +52,31 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
...
+    }
+
+    func setAyahInfo(ayahInfoData: [AyahNumber: [AyahInfo]]?) {
+        self.ayahInfoData = ayahInfoData
+        highlightingView.ayahInfoData = ayahInfoData
+    }
+
+    func highlightAyah(ayah: AyahNumber) {
+        guard let ayahInfo = ayahInfoData?[ayah] else { return }
+        if ayahInfo.count > 0 {
```
mohamede1945: As indicated in the comments. We should get ride of the following checks.
```
guard let ayahInfo = ayahInfoData?[ayah] else { return }
if ayahInfo.count > 0 {
```

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -50,5 +52,31 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
...
+            highlightingView.setScaleInfo(scale, xOffset: deltaX, yOffset: deltaY)
+        }
+    }
+
+    func setAyahInfo(ayahInfoData: [AyahNumber: [AyahInfo]]?) {
+        self.ayahInfoData = ayahInfoData
+        highlightingView.ayahInfoData = ayahInfoData
+    }
+
+    func highlightAyah(ayah: AyahNumber) {
```
mohamede1945: I think `highlightAyah` should be `highlightAyahs` that takes a `Set<AyahNumber>`. Then we don't need to have setNeedsDisplay. Because we will be setting the ayahs and the `HighlightingView` will be responsible for the best way to redraw for example if we are using core animation we would use another method.

So, I suppose at the end this method should be something like:
```
func highlightAyahs(ayahs: Set<AyahNumber>) {
    highlightingView.highlightedAyat = ayahs
}
```

Quran/HighlightingView.swift
```diff
@@ -12,7 +12,51 @@ import UIKit
 // It's also expected to reuse layers instead of dropping & creating new ones.
 class HighlightingView: UIView {
 
-    @IBInspectable var highlightColor: UIColor = UIColor.redColor()
+    @IBInspectable var highlightColor: UIColor = UIColor(red: 0.275, green: 0.58, blue: 0.651, alpha: 0.25)
 
-    var highlightableAreas: [Rect] = []
+    var highlightedAyat: [AyahNumber] = [] {
```
mohamede1945: Can we declare `highlightedAyat` to be of type `Set<AyahNumber>`

Quran/AyahInfo.swift
```diff
@@ -7,18 +7,33 @@
 //
 
 import Foundation
+import CoreGraphics
```
mohamede1945: I think we need to remove CoreGraphics from here since this is the Model layer.

Quran/AyahInfo.swift
```diff
@@ -7,18 +7,33 @@
...
     let maxX: Int
     let minY: Int
     let maxY: Int
-    var rect: Rect {
-        get {
-            return Rect(x: minX, y: minY, width: maxX - minX, height: maxY - minY)
-        }
+
+    func rect(scale: CGFloat, xOffset: CGFloat, yOffset: CGFloat) -> CGRect {
+        return CGRect(x: CGFloat(minX) * scale + xOffset,
```
mohamede1945: According to the last comment, we need to remove the scaling from here also, since it has nothing to do with the Model. I suggest do the following:
1. Revert `var rect: Rect` back again.
2. Add the scaling function as an extension over CGRect in a new file named `CGRect+Extension.swift` in Classes -> Helper -> Extensions.

ahmedre:
updated based on comments - btw, not sure if the extension makes sense for `CGRect` because it's basically a static method (and because our "source object" is a `Rect` instead of a `CGRect`), but i did it anyhow.

mohamede1945:
Thank you so much for your efforts. Jazak Allah khayran

Issue #4, opened 2016-04-18
Tafseer / Translation


mohamede1945:
- [ ] Add the ability to switch between Translation and regular reading experience with Madani Quran.
- [ ] If there are no translations downloaded yet. Show the Translation download manager.
- [ ] When switching to translation view, always show the last selected translation or the first one from the list.
- [ ] Give the user the ability to change current translation from the currently downloaded translations and add a link to download more to show the Translation download manager if not all translations downloaded yet.
- [ ] The translation view show always show the Ayah in Arabic.
- [ ] The listening experience should be the same as if the user reading Arabic Madani Quran. Highlight the current Ayah in Arabic and its translation and automatically scroll to the currently playing ayah, etc.

PR #39, opened 2016-06-20
Selecting verses

hossamghareeb:
Done adding feature of selecting verses via long press gesture. A menu will be shown with options Copy & Share the text of selected verse.


mohamede1945:
Great! Jazaka Allah khyran

I might be late in review since I'm traveling now.

mohamede1945:
That's an amazing work! Rabena yekremek. I've just some simple comments:

1. Can you fix Swiftlint violations? It seems many violations, but most of them can be corrected with "swiftlint autocorrect" command.
2. There is 2 warning regarding the Selector that suggesting to use #selector instead. Also, please fix it.
3. Can you remove the debugging `print` statements?
4. For the `ShareController`. I don't like how we loop over to get the visible controller and present the `UIActivityViewController`. I'd love to see a delegate in the  `HighlightingView` that will be called instead of `ShareController.showShareActivityWithText(text, image: nil, url: nil, handler: nil)` and in the `QuranViewController` should be the one responsible for presenting the `UIActivityViewController`.
5. Can we rename the method `shouldListenToTapGestureAtPage ` to something like `deselectSelectedVerse`. Because the current name doesn't feel like having a side effect. It feels like it is only a check, but the method has a side effect of `highlightingView.deselectTheSelectedVerse`.

But really amazing work and like how simple the solution is. May Allah accept from you.

mohamede1945:
I forgot to apologize for the late review. Sorry for that.

hossamghareeb:
Hey Mohamed, 

No need to apologize :) its fine. I apologize too for my late replay. 

Most of your points are valid and will work on it. I didn't work with SwiftLint before, so I will give it a shot. 

For point #4, the problem is that to reach `QuranViewController` from `HighlightingView` we need multiple delegates for that. I'm thinking of using NSNotificationCenter but I'm not a big fan of it. What do you think?

mohamede1945:
I also prefer using delegate over notification.

hossamghareeb:
Hey Mohamed, 

All issues should be fixed. Could you have a look?

mohamede1945:
Thank you for your fixes.
1. I still can see few swiftlint violations. 
```
/Work/OpenSource/Quran/Quran-iOS/Quran/HighlightingView.swift:100: warning: Line Length Violation: Line should be 150 characters or less: currently 174 characters (line_length)
/Work/OpenSource/Quran/Quran-iOS/Quran/HighlightingView.swift:160:14: warning: Valid Docs Violation: Documented declarations should be valid. (valid_docs)
/Work/OpenSource/Quran/Quran-iOS/Quran/HighlightingView.swift:167:5: warning: Valid Docs Violation: Documented declarations should be valid. (valid_docs)
/Work/OpenSource/Quran/Quran-iOS/Quran/ShareController/ShareController.swift:14: warning: Line Length Violation: Line should be 150 characters or less: currently 200 characters (line_length)
```

Quran.xcodeproj/project.pbxproj
```diff
@@ -1143,7 +1160,7 @@
 				TargetAttributes = {
 					54B265E21CC57D8E00D3AEF8 = {
 						CreatedOnToolsVersion = 7.3;
-						DevelopmentTeam = MKP73M5X4A;
```
mohamede1945: Can you revert all changes related to signing identities, dev team and bundle id?

Quran/HighlightingView.swift
```diff
@@ -8,12 +8,22 @@
...
 // This class is expected to be implemented using CoreAnimation with CAShapeLayers.
 // It's also expected to reuse layers instead of dropping & creating new ones.
 class HighlightingView: UIView {
 
     @IBInspectable var highlightColor: UIColor = UIColor.appIdentity().colorWithAlphaComponent(0.25)
 
+    /// is true when user long pressed on verse to select and a menu controller is displayed.
+    var isSelectingVerse = false
+
+    weak var delegate: HighlightingViewDelegate!
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`?

Quran/HighlightingView.swift
```diff
@@ -73,4 +84,123 @@ class HighlightingView: UIView {
...
+                isSelectingVerse = true
+            }
+
+            return true
+        }
+        return false
+    }
+
+
+    private func ayahNumberForTouchLocation(location: CGPoint) -> AyahNumber! {
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`?

Quran/HighlightingView.swift
```diff
@@ -8,12 +8,22 @@
 
 import UIKit
 
+@objc protocol HighlightingViewDelegate {
```
mohamede1945: Is there a specific reason to have `@objc` protocol? I think that protocol can just be constrained to be a class protocol `protocol HighlightingViewDelegate: class``.

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -11,7 +11,12 @@ import AVFoundation
 
 private let imageHeightDiff: CGFloat = 34
 
-class QuranPageCollectionViewCell: UICollectionViewCell {
+@objc protocol QuranPageCollectionCellDelegate {
```
mohamede1945: Same thing for this protocol, I think it could be just a class protocol no need for `@objc`, what do you think?

Quran/ShareController/ShareController.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Created by Hossam Ghareeb on 6/20/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import UIKit
+
+class ShareController: NSObject {
+
+
+    class func showShareActivityWithText(text: String, image: UIImage! = nil, url: NSURL! = nil, sourceViewController: UIViewController, handler: UIActivityViewControllerCompletionWithItemsHandler!) {
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`?

Quran/ShareController/ShareController.swift
```diff
@@ -0,0 +1,33 @@
...
+        itemsToShare.append(text)
+        if image != nil {
+            itemsToShare.append(image)
+        }
+        if url != nil {
+            itemsToShare.append(url)
+        }
+
+        let activityViewController = UIActivityViewController(activityItems: itemsToShare, applicationActivities: nil)
+        if let handler = handler {
```
mohamede1945: I don't think we need to if let here, you can just assign the handler directly.

mohamede1945:
The last request would be to squash your commits as few as possible.

mohamede1945:
Forgot to say. Happy Eid :)

Quran.xcodeproj/project.pbxproj
hossamghareeb: Opps, its pushed by mistake, sorry :)

Quran/HighlightingView.swift
hossamghareeb: because it will be used as a `weak` reference to avoid retain cycle. You can't mark non ObjC protocols as weak as far as I know.

Quran/QuranPageCollectionViewCell.swift
hossamghareeb: Same, we need to avoid retain cycle by marking it as weak when used.

hossamghareeb:
Happy Eid to you too 👍

Quran/HighlightingView.swift
mohamede1945: That's true, but if you need to do it just for swift, you can use `class` constraint. [Explained here](http://stackoverflow.com/questions/24066304/how-can-i-make-a-weak-protocol-reference-in-pure-swift-w-o-objc).

hossamghareeb:
Hi Mohamed, I fixed the issues and squashed commits.

Quran.xcodeproj/project.pbxproj
mohamede1945: This is still not fixed!

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -21,13 +25,16 @@ class QuranPageCollectionViewCell: UICollectionViewCell {
     @IBOutlet weak var mainImageView: UIImageView!
 
     @IBOutlet weak var scrollView: UIScrollView!
+    weak var cellDelegate: QuranPageCollectionCellDelegate!
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`.

Quran/QuranPagesDataSource.swift
```diff
@@ -18,6 +18,8 @@ class QuranPagesDataSource: BasicDataSource<QuranPage, QuranPageCollectionViewCe
 
     var highlightedAyat: Set<AyahNumber> = Set()
 
+    weak var pageCellDelegate: QuranPageCollectionCellDelegate!
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`.

Quran/ShareController/ShareController.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+
+import UIKit
+
+class ShareController: NSObject {
+
+
+    class func showShareActivityWithText(text: String, image: UIImage? = nil, url: NSURL? = nil,
+                                         sourceViewController: UIViewController,
+                                         handler: UIActivityViewControllerCompletionWithItemsHandler!) {
```
mohamede1945: Please use `Optional` instead of `ImplicitlyUnwrappedOptional`.

hossamghareeb:
Done :) 

Sorry for my late reply

mohamede1945:
Thank you so much for this great effort!

Issue #41, opened 2016-07-11
Pause/Play doesnt work on first time


waseem01:
1. Launch app
2. Tap on Surah Baqarah
3. Tap play button at the bottom.
4. Observe, button first shows play mode and then changes to pause mode
5. Tap the button, the recitation doesnt stop, tap again, now it stops
6. This happens only first time for any surah when its loaded.

waseem01:
I will fix this inshaAllah, @mohamede1945 please add me to the project so I can assign it to myself

ahmedre:
@waseem01 @mohamede1945 i've added an "in progress" label that we can label things people are working on with instead of adding to the project.

waseem01:
@ahmedre Thanks, I dont have 'write permissions' to the repo to move it :)

ahmedre:
we're not adding people with write permissions to the project right now for a few reasons:
1. as a "github organization," we pay depending on the members in the organization
2. we want to restrict checkin permissions to the code so only approved pull requests get merged

i also looked to see if there's a way to just give permissions to add labels and what not, but i didn't see that, though i could have missed it.

ahmedre:
as an aside, most open source projects work this way - i.e. you don't get write permissions to most of the repositories to be able to contribute - the maintainers are responsible for looking at pull requests, reviewing them, and merging code accordingly.

ahmedre:
oh - and i forgot, most importantly - jazakAllah khairan in advance!

waseem01:
Thanks, I will open a PR and I guess the admin can merge.

mohamede1945:
@waseem01 I don't know why I'm not able to reproduce the issue. Can you have tell me which reciter were you using?

mohamede1945:
@ahmedre Thanks a lot for creating the label. I was mainly depending on zenhub for such things, but I think I should use labels too since it will be easier for everyone.

mohamede1945:
Gif of the issue:
![g7jxoq](https://cloud.githubusercontent.com/assets/20160610/16753775/327413c8-47b2-11e6-915d-8f9f1a88b37f.gif)

mohamede1945:
I'll start working on this issue now.

mohamede1945:
Can't reproduce anymore.

PR #42, opened 2016-07-11
Fix play/pause on audio played first time

waseem01:
https://github.com/quran/quran-ios/issues/41

I am not for removing the line of code, however I did see removing it.
1. Doesn't change existing functionality 
2. Fixes the issue

onPlaybackPaused was getting called right after onPlaybackResumed because the deleted line of code.

### DIFF
--- a/Quran/QueuePlayer.swift
+++ b/Quran/QueuePlayer.swift
@@ -120,7 +120,6 @@ class QueuePlayer: NSObject {
 
         rateObserver = observe(retainedObservable: player, keyPath: "rate", options: [.New]) { [weak self] (observable, change: ChangeData<Float>) in
             self?.updatePlayNowInfo()
-            self?.onPlaybackRateChanged?(playing: change.newValue != 0)
         }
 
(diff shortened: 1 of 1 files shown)


mohamede1945:
Thanks a lot for your contribution! But This actually introduces another problem. Currently, the KVO observer is used because the iOS could forcibly pause the player in some situations (e.g. a phone call). The code you commented will update the UI to reflect that. So, I don't think this is the way to go.

But thank you so so much for your help. I would encourage you to keep digging to find another solution.

Jazaka Allah Khayran

waseem01:
@mohamede1945 aha that make sense, I was wondering when the call was happening.
g7jxoq

mohamede1945:
Thanks a lot for that! I also hate KVO, especially if you are writing Swift. But this is the only way to know if the system pauses the audio player.

mohamede1945:
Closing the PR! Since it doesn't resolve the issue https://github.com/quran/quran-ios/issues/43

PR #45, opened 2016-10-25
Migrate to Swift 3

ahmedre:
This is a work in progress patch to use Swift 3.

Status:
- compiles without warnings or errors
- runs and can read just fine

What's left:
- fix crashes related to downloading stuff (audio, etc)
- fix "copy" functionality recently added


mohamede1945:
That's amazing! Thank you so so much! Jazaka Allah khyran!

I'll keep this PR open until we completes what's left. When I start working inshaa' Allah next week I'll try to do what's left, will let you know beforehand.

ahmedre:
waiyyakum - the above issues are fixed and everything works. for ease of reviewing, look at the commits (most of the line counts come from the first commit, which is updating Pods, and for that one, you just need to look at the `Podfile`).

i made some decisions that i hope you will have better solutions for, so i will document these in the code, but i want to document here the three things i did that bother me the most:

1. removed `Cache` abstraction for now (use `NSCache` type directly for the image cache) - since `NSCache` now accepts generics, this causes problems with the protocol (which can't take generics), and i couldn't make things play well together.
2. fixing the broken downloading issue mentioned above - the root cause of this happens to be that `URLRequest` lookups into the `downloadRequests` dictionary were failing. i have no idea why (i can verify by looking at a combination of the `url` field on the `request` we are doing the lookup by and the dictionary itself that it exists) - see [the 4th commit](https://github.com/quran/quran-ios/pull/45/commits/c9a35e097b177665c804a3fb92073830f30ab568). i changed this to key off of the `URL` directly instead of off the `URLRequest`.
3. fixing copy - see [the 5th commit](https://github.com/quran/quran-ios/pull/45/commits/397b6ae8c0806f291d46e88744c7c1c9c9bf32b4). i basically added an item to the `UIMenuController` for copy to be able to handle the action - modifying the existing copy method declaration didn't work because it caused a conflict with the version from `UIResponder`. i am guessing there must be a better way to fix this.

i'll go ahead and add comments inside the PR for other places i want to draw your attention to as well.

jazakumAllah khairan!

Quran/QuranURLs.swift
```diff
@@ -0,0 +1,16 @@
...
+//  URL.swift
+//  Quran
+//
+//  Created by Mohamed Afifi on 5/2/16.
+//  Copyright © 2016 Quran.com. All rights reserved.
+//
+
+import Foundation
+
+struct QuranURLs {
```
ahmedre: i renamed this from `URL` due to us now having `Foundation.URL` instead of `NSURL`.

Quran/QuranPageCollectionViewCell.swift
```diff
@@ -75,47 +75,47 @@ class QuranPageCollectionViewCell: UICollectionViewCell, HighlightingViewDelegat
...
+    fileprivate func scrollToHighlightedAyat() {
         guard let first = highlightingView.highlightingRectangles.first else {
             return
         }
 
         layoutIfNeeded()
 
         var union = first
-        highlightingView.highlightingRectangles.forEach { union.unionInPlace($0) }
+        highlightingView.highlightingRectangles.forEach { union = union.union($0) }
```
ahmedre: i couldn't seem to find `unionInPlace` so ended up doing this instead.

Quran/Crash.swift
```diff
@@ -45,24 +45,22 @@ struct Crash {
             instance.setBoolValue(value, forKey: key.key)
         } else if let value = value as? CustomStringConvertible {
             instance.setObjectValue(value.description, forKey: key.key)
-        } else if let value = value as? AnyObject {
```
ahmedre: i removed this because there's no `description` on `AnyObject`

Quran/UIImage+Extension.swift
```diff
@@ -12,25 +12,29 @@ extension UIImage {
 
     func preloadedImage() -> UIImage {
 
+        guard let cgimg = cgImage else {
```
ahmedre: i added this because `cgImage` is now `cgImage?`

Quran/UIColor+Extension.swift
```diff
@@ -29,7 +29,7 @@ extension UIColor {
 
      - returns: The 1 pixel image of the color.
      */
-    func image() -> UIImage {
+    func image() -> UIImage? {
```
ahmedre: `UIGraphicsGetImageFromCurrentImageContext` now returns an optional

Quran/QuranURLs.swift
mohamede1945: That's Ok.

Quran/QuranPageCollectionViewCell.swift
mohamede1945: That's Ok.

Quran/UIImage+Extension.swift
mohamede1945: That's Ok :)

Quran/UIColor+Extension.swift
mohamede1945: That's Ok :)

mohamede1945:
Really really thank you so much for your work! It took from us more than a week to migrate our app. So, I really know how much work you did in this PR.

1. Regarding the pods, you can upgrade to cocoapods 1.1.x that supports swift 3, but the loop is the workaround for previous versions which is just perfect for backward compatibility.

2. Regarding the `NSCache`, I remember the problem with me using `NSCache` is that it crashes the app with memory exception, that was weird for me since it is its only job, right? Anyway, I will test it and see if the [new swift implementation](https://github.com/apple/swift-corelibs-foundation/blob/19249417b01573bd6aa32b9a24cc42273315a48b/Foundation/NSCache.swift) handles this better :)

3. Regarding the fix for the download, that's very good. I'll have to actually use a different persistence store than `NSUserDefaults`.

4. That's good too since it does the job.


Really hats off for this PR.  May Allah accept from you.

ahmedre:
waiyyakum, and jazakAllah khairan also!

PR #52, opened 2016-11-03
I have added Dutch 

abdelouali:


mohamede1945:
Jazaka Allah Khyran! Thank you so much! I'll wait until I go home to add it because we need to add the localizations to the Xcode project.

PR #53, opened 2016-11-03
Fix warnings and a typo in class name

ahmedre:


mohamede1945:
Jazaka Allah khyran! Nice catch for the typo :)
It seems you updated to Xcode 8.1 :) I'm on my way updating today.

PR #56, opened 2016-11-03
Use zopfli to compress the Quran 1280 width images

ahmedre:
Using zopfli compression gives pretty huge wins - the images go from 97mb
to 71mb with no difference in quality. Since these are the images we
bundle with the app, this will reduce the ipa size by ~25mb. See also
quran/quran_android#699.


ahmedre:
special thanks to @tarekeldeeb for this excellent suggestion.

@mohamede1945 i am running this for 1920 images now as well - when am done, should i commit this update, or should we just remove them from the repository?

mohamede1945:
Seems amazing! Thank you so much!

Regarding the 1920, I think we should delete them from the repo since they are not used, I totally forgot about them :)

mohamede1945:
Thank you @tarekeldeeb for the suggestion :)

@ahmedre I read the other thread and I saw you talking about 100 iterations instead of 500. Since for us here, every byte counts? If the overall decrease (between the 100 and 500) is 1mb then that means I have space for like 10 icons and images to add in the app :)

ahmedre:
the difference is much smaller than that - 14 bytes for page 300. if we assume an average of 20 bytes per page, that's 20 * 604 = 12080 / 1024 = 11.79kb savings across all the images.

consider the time - on my machine, 46.5 seconds for 100 iterations, 3 minutes and 8 seconds for 500 iterations - so ~7.8 hours for 100 passes, versus ~31.5 hours.

if you feel the 12kb is worth it, i'll re-run it :)

mohamede1945:
Oh 31 hours 🙀! Never mind 😂

tarekeldeeb:
@ahmedre @mohamede1945 Thanks a lot guys for your enthusiasm.

PR #59, opened 2016-11-05
History of 3 last pages

### LINKED ISSUE #54: History Of Last Pages
* Add the ability to have multiple last pages (3 for now). Each item represents a session.
* The user can resume a session if he goes to Bookmarks/Last Page and tap one of them
* The user can create a new session by opening the Quran from any other place (e.g. Juz tab, Sura tab, Ayah Bookmark, Page Bookmark).
* History of last pages shall be sorted by last update date.

mohamede1945:
Fixes #54


Quran/SqliteLastPagesPersistence.swift
```diff
@@ -0,0 +1,111 @@
...
+        return run { connection in
+            let insert = LastPages.table.insert(
+                LastPages.page <- page,
+                LastPages.createdOn <- Date(),
+                LastPages.modifiedOn <- Date())
+            let rowId = try connection.run(insert)
+            let rows = try connection.prepare(LastPages.table.filter(LastPages.id == rowId))
+
+            // keep top maxNumberOfLastPages rows only
+            let statement = "DELETE FROM last_page WHERE _ID NOT IN (SELECT _ID FROM last_page ORDER BY modified_date DESC LIMIT \(maxNumberOfLastPages))" // swiftlint:disable:this line_length
```
mohamede1945: Wasn't able to come up with Sqlite.Swift way to generate this statement.

ahmedre: if you know how many history rows there are, you can do something like:
`DELETE FROM last_page ORDER BY modified_date ASC LIMIT ?`, and use `number of rows - maxNumberOfLastPages` as that value - that would avoid the subquery.

otherwise, i think it's fine :)

ahmedre:
can we also avoid duplicate pages?

simulator screen shot nov 6 2016 12 43 44 pm

ahmedre:
fixed the duplicates bug in #60

PR #60, opened 2016-11-06
Fix duplicate last pages

ahmedre:
Fix duplicate pages by making pages unique, and by using `replace into` instead of
`insert into`.


ahmedre:
note that this requires you to uninstall and reinstall the app to re-generate the last pages database (not worth an upgrade for everyone since probably no one besides you and me have this change locally).

mohamede1945:
I agree with you on that. I will next add the ability for the user to swipe to delete any bookmark (Last pages, page bookmark or ayah bookmarks) from bookmarks controller.

PR #63, opened 2016-11-08
Fix for the updating last pages

### LINKED ISSUE #61: Crash for Updating Last page from Last Pages Bookmarks
After applying https://github.com/quran/quran-ios/pull/60 now there is a crash when the user opens the Quran from one of the last pages.

Steps to reproduce:
1. Assume you have in the "last pages bookmarks", pages 1 and 2.
2. Open the Quran from page 1.
3. Then swipe the page to go to page 2. The app will crash at this point.

I think the crash in function 
```swift
func update(page: LastPage, toPage newPage: Int) throws -> LastPage
```

mohamede1945:
Fixes #61


ahmedre:
jazakAllah khairan for fixing and sorry for missing this!

PR #69, opened 2016-11-24
Added selected tab bar icons

mohamede1945:
Since I just learned some basics in sketch, I gave it a try to produce those icons :)


abdelouali:
Nice! Good job. JazakaAllah

ahmedre:
awesome jazakAllah khair!

Issue #64, opened 2016-11-08
Sometimes stopping download & cannot resume it again


mohamede1945:
Sometimes, if I stopped the download in the middle, I'm not able to resume it again.

wahyusumartha:
Hey @mohamede1945 

Thanks for the great work of Quran iOS App, I really appreciate your work, 

i was playing around with the code base, and found out that the issue happened on iOS 10 (was using simulator 10.1). It was working fine on iOS 9.3 (on my simulator) 

I was trying to patch the bug with the discussion that was going on in stack overflow. 

Hopefully that will fix the problem,

Thanks

PR #71, opened 2017-01-03
Fix issue when stopping download cannot resume it again #64

wahyusumartha:
Hey there, 

i was playing around with the code base, and found out that the issue happened on iOS 10 (was using simulator 10.1). It was working fine on iOS 9.3 (on my simulator)

Hopefully that will fix the problem,

Thanks


mohamede1945:
Oh, So sorry for late reply. Thank you so much for the fix!  Jazaka Allah khyran!

Can you explain to me the approach you took in fixing it?

Because as far as I know the problem was because Swift 2.2 serialize dictionaries in a way that's different than Swift 3.0. Basically, if a user has the app installed before with Swift 2.2 and have an unfinished download, then he upgraded, he will face this problem. But since I already shipped the app with Swift 3.0 and most of our users upgraded already, I don't think we need a fix it anymore.

So, probably your PR might be fixing a different bug that I don't know it.

Actually, the current way of using `UserDefaults` is wrong since it is not mean to save this type of data. We should use a different persistence like sqlite or core data.

mohamede1945:
Closing it since it's downloading persistence is using sqlite not UserDefaults anymore.

PR #74, opened 2017-02-11
Fix lint issues so app can compile

ahmedre:


ahmedre:
this PR contains 4 commits:
the first commit (0da2e4b) isn't really interesting - it's mostly the result of `swiftlint autocorrect` and some correction of things that autocorrect missed or that it got wrong.

would appreciate a double check on the other 3 commits though - the second and third one replace tuples with structs for some tuples with more than 2 arguments - the last one ignores a lint warning that, imho, doesn't apply in this case.

mohamede1945:
It seems perfect to me. Jazaka Allah Khayran

PR #76, opened 2017-02-14
WIP: Translations

ahmedre:
This is a work in progress diff for implementing translations within Quran for iOS. I'll update when it's ready for review.


mohamede1945:
That's a perfect work! Jazaka Allah Khyran for it. 

I just have a small objection to having the title part of the back button. I don't have any problem having the title left aligned but not so close to the back button (e.g. the latest updates from WhatsApp has the title left aligned). I looked at the Calendar app, but it has the back title properly reference the previous screen.

ahmedre:
sounds good - out of curiosity, will you just do this by changing the title view's rect?

mohamede1945:
As far as I remember, I need to set a custom view so I'll use my own `UILabel`.

PR #77, opened 2017-02-19
SQLite Downloads Persistence

mohamede1945:
Advantages of this PR.

1. Using more reliable persistence storage (sqlite) instead of NSUserDefaults with the previous implementation depending on implementation details of Swift.Dictionary type.

2. Fixes an issue when a download is canceled, but the progress bar doesn't disappear.

3. Orienting the Downloader into working with download batches so that user can download multiple batches at the same time (e.g. Audio and Translations).

4. Fixes https://github.com/quran/quran-ios/issues/64


ahmedre:
jazakAllah khairan, awesome!

PR #82, opened 2017-04-02
Change project license to GPLv3.

mohamede1945:
Still, need to update all source code files to add a header with a short license description.
Also, we need to add a Swiftlint rule file_header.


abdelouali:
Hi Mohamed, If I can do something without the mac, i'am still here :)

ahmedre:
jazakAllah khairan

PR #87, opened 2017-04-08
Including GPLv3 summary to all files

mohamede1945:
Fixes https://github.com/quran/quran-ios/issues/81

I also added some opt-in rules from swiftlint that have been introduced in newer versions of the tool. I'm using the latest 1.8.1 version.


mohamede1945:
Not needed, it is included in #88

PR #122, opened 2017-04-14
Fixes a crash when user tries to share a verse on iPad devices

### LINKED ISSUE #121: App crashes While Sharing on iPad
Long tap on a verse, then choose "share.." on an any iPad device.

mohamede1945:
Fixes #121


mohamede1945:
I will merge it myself, as I want to release 1.2.1 tomorrow inshaa'Allah.

PR #126, opened 2017-04-15
Better Error handling

### LINKED ISSUE #121: App crashes While Sharing on iPad

### LINKED ISSUE #123: Recovery From Corrupted Downloaded Translation File
Sometimes the app downloads a corrupted translation file. It would be great if we tracked down the cause of the problem, but as a quick fix, add a recovery from this error otherwise the app is not usable at all.

mohamede1945:
Fixes #121 
Fixes #123 by both preventing downloading the file if there is a server error (i.e. unacceptable HTTP status code) and also the recovery of bad translations SQLite files.

Also using `DispatchQueue.default` so that it is easy to debug Promises just by modifying `DispatchQueue.default` in the `AppDelegate` to be `zalgo`.

### DIFF
--- a/Pods/GenericDataSources/Sources/DataSourceSelector.swift
+++ b/Pods/GenericDataSources/Sources/DataSourceSelector.swift
@@ -29,12 +29,14 @@ import Foundation
     case move
     case willDisplayCell
     case didEndDisplayingCell
-    case shouldShowMenuForItemAt
     case canPerformAction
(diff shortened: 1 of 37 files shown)


mohamede1945:
I will merge it myself, as I want to release 1.2.1 tomorrow inshaa'Allah.

PR #128, opened 2017-04-16
Using scrollViewWillBeginDragging to dismiss the bars

### LINKED ISSUE #110: Dismiss The Bars When User Starts Scrolling
iBook makes it easier to dismiss the bars for example when the user starts scrolling the bars are dismissed.

mohamede1945:
Fixes #110 
And also setting `ITSAppUsesNonExemptEncryption` to `false` in Info.plist so that we are not asked about it every time we submit a release.


ahmedre:
jazakAllah khairan

PR #129, opened 2017-04-16
Fixes for TabBar scrolling & Edit/Done behavior for Translations

### LINKED ISSUE #106: Tapping on the Tabbar should scroll correctly
**From collin,**

- Tapping the tab buttons on the home screen when the tab is already active causes the first item on the list to scroll out of view.

### LINKED ISSUE #112: Correct Edit/Done Behavior for Translations
When the user swipes to delete the **Edit** bar button should be converted to **Done** and when the user taps it, the delete button should disappear.

mohamede1945:
Fixes #106 & Fixes #112

### DIFF
--- a/Quran/BaseNavigationController.swift
+++ b/Quran/BaseNavigationController.swift
@@ -20,10 +20,14 @@
 
 import UIKit
 
-class BaseNavigationController: UINavigationController {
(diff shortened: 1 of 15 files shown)


ahmedre:
jazakAllah khairan

PR #137, opened 2017-05-01
Audio Download Manager

### LINKED ISSUE #31: Audio Downloads Tab
Implementing Audio Downloads Tab.
- [ ] Should show Reciters and how many suras are downloaded.
- [ ] User can delete downloads of a particular reciter with a confirmation.
- [ ] User can tap on the reciter to see more details.
- [ ] The details should show all suras and next to each sura a delete (if downloaded) or a download (if not) icon or progress (if download in progress). 
- [ ] Tapping the sura line should perform the action delete/download/cancel download.
- [ ] There should be a button to download all suras, hide that button if all suras are already downloaded.

mohamede1945:
Fixes #31 

This PR mainly supports Audio Download Manager.
It also includes modularizing the app into multiple framework targets (`BatchDownloader`, `SQLitePersistence` for common functionality, `VFoundation`, `UIKitExtension`, `QueuePlayer`). I'm planning to separate those frameworks into their own repo later specifically `BatchDownloader` and later `QueuePlayer` (when we implement verse repetition) for their importance.

In this PR, I had enhanced `BatchDownloader` to support downloading any number of files. Previously it was taking more than 1 min to just prepare (download not event started) to download the entire gapped Quran (6,236 files) and besides most of the files fail with timeout. But with the current version, we can enqueue almost all files for all reciters and the downloader saves them in SQLite persistence and downloads only 200 (configurable) files at a time.

Besides, some other enhancements explained in the commit messages.


mmahalwy:
🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

PR #140, opened 2017-05-02
Fix for Audio Download Screen

mohamede1945:
* Fix for an issue that makes the downloading progress always visible even after the downloading is completed.
* Increase frequency of reloading the downloaded size to be every 0.05% (around 1MB) previously was every 1%.


mohamede1945:
I found some small issues with the audio downloader, I will fix them and re-open the PR again.

PR #141, opened 2017-05-06
Fixes & Performance Improvements for Audio Download Manager

mohamede1945:
* Increased performance of loading audio manager tab from 7-8 seconds to 1-2 seconds.
* Increased max simultaneous downloads to 600 so any Juz' can be downloaded on a one shot.
* Fix an issue for reloading a different cell than the downloading cell.

They are mostly independent fixes, so individual commits might explain it.


ahmedre:
jazakAllah khairan

yshalabi:
Wa iyyakum, husn Al jazaa. Thank you so much for open sourcing it!! I think
GPL is such a great choice.

Btw I'll be passing through San Francisco in a few weeks. It'll be nice to
meet anyone who might live there.

Yasser

PR #142, opened 2017-05-06
Update QariAudioFileListRetrieval.swift

yshalabi:
build error fix: cast dbFile to base type QariAudioFile to allow using + operator to combine the QariSuraAudioFile list and the single QariDatabaseAudioFile (dbFile).


mohamede1945:
I don't have any build errors with this line of code. Can you tell me which Xcode version are you using?

mohamede1945:
This is a problem with older Xcode, if you use the latest Xcode and Swift, you shouldn't have this compiler error see discussion #149.

PR #160, opened 2017-05-22
Admob

tuanquanghpvn:


mohamede1945:
Salam Alikom, just wondering what is the purpose of implementing this?

abunur:
I can't imagine anything worse than a Quran app with ads.  Especially random stuff you can't control, could be _so_ inappropriate.

Abdullah076:
agree with brother abu nur. seeing ads in Quran or any other islamic apps are just disgusting to me. as many times inappropiate contents show-up.

if someone wants to get donation, he may add option to donate in their app. that is ok. but ADDING ADS, i cant agree with.

mohamede1945:
Closing it, as we will never display Ads in this application.

PR #161, opened 2017-05-22
Ability to Search in Arabic Quran & Translations

### LINKED ISSUE #3: Search
* Add ability to search for text in Ayah.
* Add ability to search for text in downloaded translations.

mohamede1945:
Fixes #3 

The PR includes the ability to search for specific queries in Quran & the translations.
I integrated `RxSwift` to implement this feature (for the business logic part, UI still using the old `UIKit` way), I also tried to use `VIPER` architecture, which I think is a good architecture.

`TOOD:` after implementation of VIPER, I think at least I should create Routers (or others call it Coordinators) for existing MVC modules as it simplifies the ViewController code.

Also, there are some places where I have `// Design:` comment that I should fix when creating the routers for other modules.


mohamede1945:
If there are no objections, I will merge it in 30 mins as I want to upload a version to `TestFlight` before I sleep.

PR #183, opened 2017-06-10
pod install, Swiftlint

abunur:
-added Swiftlint to Podfile
-rest of changes are from pod install


mohamede1945:
Looks great! Jazak Allah khyran! The only thing missing is to replace the call in the Run Script.

abunur:
Ok will add that in next pull request, jzk

PR #189, opened 2017-06-14
gitattributes

abunur:
added .gitattributes file
makes merging storyboards easier


mohamede1945:
Sounds great! Jazak Allah khyran

PR #192, opened 2017-06-17
SwiftLint

abunur:
invocation for SwiftLint added to script build phase


mohamede1945:
Thanks for creating this PR. But I don't think we should modify `[CP] Embed Pods Frameworks` build script because it is automatically created by cocoapod. Instead, just modify `Swiftlint Run Script`.

PR #225, opened 2018-04-11
Added review controller

### LINKED ISSUE #214: SKStoreReviewController - asking for rating
It would be nice to add SKStoreReviewController in order to increase number of people giving rating to the app.

I believe it would increase position in app store search for Quran apps. As a result more people would be able to use this ad free and so useful  app!

mumer92:
Check for app install date and number of counts app opened, then ask for Review and Rating.
This should close #214

### DIFF
--- a/Quran/Review.swift
+++ b/Quran/Review.swift
@@ -38,13 +38,11 @@ struct ReviewService {
             let appInstalledDate = simplePersistence.valueForKey(.appInstalledDate)
             let oldDate = Date(timeIntervalSince1970: appInstalledDate)
             let today = Date()
-            let difference = Calendar.current.dateComponents([.day, .minute], from: oldDate, to: today)
+            let difference = Calendar.current.dateComponents([.day], from: oldDate, to: today)
 
             if let daysCount = difference.day {
-                if daysCount > 10 {
-                    if appOpenedCounter > 7 {
-                        requestReview()
-                    }
+                if daysCount > 7 && appOpenedCounter > 10 {
+                    requestReview()
                 }
             }
         }
--- a/Quran/Review.swift
+++ b/Quran/Review.swift
@@ -35,14 +35,20 @@ struct ReviewService {
             let date = Date().timeIntervalSince1970
             simplePersistence.setValue(date, forKey: .appInstalledDate)
         } else {
-            let appInstalledDate = simplePersistence.valueForKey(.appInstalledDate)
-            let oldDate = Date(timeIntervalSince1970: appInstalledDate)
-            let today = Date()
-            let difference = Calendar.current.dateComponents([.day], from: oldDate, to: today)
+            let alreadyAskedForReview = simplePersistence.valueForKey(.alreadyAskedForReview)
 
-            if let daysCount = difference.day {
-                if daysCount > 7 && appOpenedCounter > 10 {
-                    requestReview()
+            if !alreadyAskedForReview {
+                let appInstalledDate = simplePersistence.valueForKey(.appInstalledDate)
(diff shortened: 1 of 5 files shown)


mohamede1945:
Wow, that was pretty fast. Mashaa'Allah :)

Quran/Review.swift
```diff
@@ -0,0 +1,61 @@
...
+            simplePersistence.setValue(date, forKey: .appInstalledDate)
+        } else {
+            let appInstalledDate = simplePersistence.valueForKey(.appInstalledDate)
+            let oldDate = Date(timeIntervalSince1970: appInstalledDate)
+            let today = Date()
+            let difference = Calendar.current.dateComponents([.day, .minute], from: oldDate, to: today)
+
+            if let daysCount = difference.day {
+                if daysCount > 10 {
+                    if appOpenedCounter > 7 {
```
mohamede1945: nit: Can we combine this statement and previous one into one if condition? like this `daysCount > 7 && appOpenedCounter & 10`
Also, I think the numbers are swapped.

mumer92:
@mohamede1945 Oh I missed that, now I have updated the logic. 
```
if let daysCount = difference.day {
    if daysCount > 7 && appOpenedCounter > 10 {
        requestReview()
    }
}
```

mumer92:
As for the review process, according to the documentation: [https://developer.apple.com/app-store/ratings-and-reviews/](url)

> Although you should call this method when it makes sense in the user experience flow of your app, the actual display of a rating/review request view is governed by App Store policy. Because this method may or may not present an alert, it's not appropriate to call it in response to a button tap or other user action.

So we cannot know if user has actually submitted a review or cancelled the dialog. 

As for the limit of app asking user for the review, it is `3 times / 265 days`, according to: [https://developer.apple.com/ios/human-interface-guidelines/system-capabilities/ratings-and-reviews/](url)

> The system offers a consistent, nonintrusive way for apps to request ratings and reviews. To use this feature, you simply identify places in your app's user experience where it makes sense to ask for feedback. If the user hasn't already given feedback and a request hasn't been made too recently, the system displays an in-app prompt that asks for a rating and an optional written review. The user can supply feedback or dismiss the prompt with a single tap. (In Settings, the user can also opt out of receiving these rating prompts for all apps they have installed.) **The system automatically limits the display of the prompt to three occurrences per app within a 365-day period**.

Thus the system should automatically handle appropriate time of showing Review Controller, even if we call `SKStoreReviewController.requestReview()` repeatedly.

Quran/Review.swift
```diff
@@ -35,14 +35,20 @@ struct ReviewService {
...
+                let oldDate = Date(timeIntervalSince1970: appInstalledDate)
+                let today = Date()
+                let difference = Calendar.current.dateComponents([.day], from: oldDate, to: today)
+
+                if let daysCount = difference.day {
+                    if daysCount > 7 && appOpenedCounter > 10 {
+                        requestReview()
+
+                        simplePersistence.setValue(true, forKey: .alreadyAskedForReview)
+                    }
```
mumer92: This should now limit asking for review to one time.

mohamede1945:
Thank you very much for applying the changes and creating the PR.

Jazak Allah Khyran

mohamede1945:
If you would like to implement https://github.com/quran/quran-ios/issues/226 it is related to rating as well. 
I'm sorry I'm very busy improving the audio playback, I've a WIP PR.

mumer92:
Shukran,

Yes, I actually looked into implementing that. I have a couple of things to ask. I will start discussion on issue.

PR #227, opened 2018-04-13
Add Deeplinking of app to AppStore to write review

### LINKED ISSUE #226: Persistent App Rating Link in Settings
This issue https://github.com/quran/quran-ios/issues/214 gives us a proactive way to ask the user to review the app. But here we would like to give a persistent way for the user to navigate to the app store to rate the app.

Users would expect to have an entry in the settings screen to navigate to the app store page to write a review about the app. Similar to how it is suggested [here at the end of the page](https://developer.apple.com/documentation/storekit/skstorereviewcontroller/2851536-requestreview).

mumer92:
Added cell in Settings that opens a deeplink of app to AppStore to write review of App.
This should close #226


Quran/SettingsViewController.swift
```diff
@@ -36,6 +36,7 @@ class SettingsViewController: BaseTableBasedViewController {
             guard let `self` = self else { return }
             let item = ds.item(at: indexPath)
             item.onSelection(self)
+            self.tableView.deselectRow(at: indexPath, animated: true)
```
mumer92: I have added this line as when app navigates back from Appstore, the cell remains highlighted. 
To remove highlighted state, I added this line.

Quran/Base.lproj/Android.strings
```diff
@@ -278,3 +278,4 @@ Please note that we receive lots of emails, so we may not be able to reply to al
 "bookmarks_list_empty" = "No Bookmarks";
 "tags_list_empty" = "No Tags";
 "audio_manager" = "Audio Manager";
+"write_review" = "Write App Store Review";
```
mohamede1945: Can you use `Localizable.strings` file instead of `Android.strings`, since `Android.strings` is the translations coming from the Android app.

Quran/SettingsViewController.swift
mohamede1945: That looks perfect, thank you.

mohamede1945:
Overall looks perfect. Just change the localization file and we should be able to merge it. 
Jazak Allah khyran

mumer92:
Updated. Moved `write_review ` to `Localizable.strings`.

mohamede1945:
Jazak Allah Khyran!

PR #256, opened 2018-09-05
Improve autocomplete results

### LINKED ISSUE #169: Search - reference the ayah
When searching, the search results list should show the surah:ayah numbers to the side of each result

img_1517

mohamede1945:
Fixes #169

*  Showing just a handful of words for autocomplete result instead of a full verse text.
* Ability to show multiple results from the same verse.
* Always show search term as first autocomplete result.
* Add a search icon on the left of each result, similar to how the new app store is doing it.

The updated look for the autocompletes
simulator screen shot - ipad 5th generation - 2018-09-05 at 22 43 46


mmahalwy:
Excited for these updates :)

mohamede1945:
Thanks :)

Issue #211, opened 2017-11-04
A Better Way to Show Word-By-Word Translation


mohamede1945:
Figure out a better way to show word-by-word translation. Since currently, it is hard to see the original word as the finger hides it.

PR #286, opened 2018-09-26
Fix for #211

aboualiaa:
display the arabic text and the translation/transliteration side by side in the popup

### DIFF
--- a/Quran/SQLiteAyahTextPersistence.swift
+++ b/Quran/SQLiteAyahTextPersistence.swift
@@ -93,7 +93,7 @@ class SQLiteArabicTextPersistence: AyahTextPersistence, WordByWordTranslationPer
             case .transliteration: text = transliteration
             }
             let query = table
-                .select(text, wordType)
+                .select(text, textMadani, wordType)
                 .filter(
                     Columns.sura == position.ayah.sura &&
                     Columns.ayah == position.ayah.ayah &&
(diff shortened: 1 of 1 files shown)


mohamede1945:
Thank you so much for the fix. But I think this fix could introduce another issue.
The main target users for word translation is non-Arabic speakers where people would complain about the tashkeel of the words because they are not used to it. We already have people complaining about the translation view tashkeel and that's why I created https://github.com/quran/quran-ios/wiki/Kasra-In-Translation-View-of-The-app and I'm already trying to figure out other ways to show the same Quran text in both Arabic & Translation views.

Your idea is still great. But I think instead of showing the plain text, cut the text from the Quran image and show it. It might not be the best idea either. So, I'm open to other suggestions.

aboualiaa:
Thank you for the feedback.. I'l see what I can do

mohamede1945:
Ideally, we could mimic how the iOS doing when the user tries to move the caret similar to 

6-things-ios-to-android-switchers-will-miss-5-things-ios-to-android-switchers-will-miss-magnifying-glass-for-cursor_4-100688054-orig

aboualiaa:
I thought about that. But this approach would still have the problem where the finger covers the popover (when the user is touching on the first line). So I was thinking, maybe we should do something along the lines of this:

img_9028 trim

aboualiaa:
Sorry. I guess I should have been more verbose.

This is a cropping functionality in a document scanning app. The user can drag the corners of the selection so that they lie exactly on the document's corners. While the user drags, the app shows a magnified popover of the the area under the user's finger. This popover positions itself on the screen so that it doesn't get covered by the the user's finger (or hand). You can see while I'm dragging the lower corners, the popover is on the top of the screen. When I drag the upper corners, it stays on the top of the screen (because the bottom half of the screen is naturally covered by my hand) and it simply get's out of my finger's way by going to the opposing screen side (right or left).

demo

mohamede1945:
Yes, that is a pretty good idea.

mohamede1945:
Oh sorry, the PR is closed because I deleted `xcode10` branch. You can create your PR against master instead.

PR #318, opened 2019-04-29
Improve Arabic search

ahmedre:
This patch improves Arabic search by fuzzily replacing certain
characters with a set of other acceptable ones (ex ا should match أ, إ,
etc). This works by switching the Arabic SQL query with a like query,
replacing these characters with an _. It then postprocesses them,
filtering out the ones that match the pattern with a character that is
unrelated to the one that was searched with.

### DIFF
--- a/Quran/SQLiteSearchableAyahTextPersistence.swift
+++ b/Quran/SQLiteSearchableAyahTextPersistence.swift
@@ -22,6 +22,36 @@ private struct Database {
 class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, SearchableAyahTextPersistence {
 
     let filePath: String
+    let regex = "[\u{0627}\u{0623}\u{0621}\u{062a}\u{0629}\u{0647}\u{0649}]"
+    let replacements: [Character: String] = [
+        // given: ا
+        // match: آأإاﻯ
+        "\u{0627}" : "\u{0622}\u{0623}\u{0625}\u{0627}\u{0649}",
+
+        // given: ﺃ
+        // match: ﺃﺀﺆﺋ
+        "\u{0623}" : "\u{0621}\u{0623}\u{0624}\u{0626}",
+
+        // given: ﺀ
+        // match: ﺀﺃﺆ
+        "\u{0621}" : "\u{0621}\u{0623}\u{0624}\u{0626}",
+
+        // given: ﺕ
+        // match: ﺕﺓ
+        "\u{062a}" : "\u{062a}\u{0629}",
+
+        // given: ﺓ
+        // match: ﺓتﻫ
+        "\u{0629}" : "\u{0629}\u{062a}\u{0647}",
+
+        // given: ه
+        // match: ةه
+        "\u{0647}" : "\u{0647}\u{0629}",
+
+        // given: ﻯ
+        // match: ﻯي
+        "\u{0649}" : "\u{0649}\u{064a}"
+    ]
 
     init(filePath: String) {
         self.filePath = filePath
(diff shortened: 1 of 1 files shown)


Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -39,22 +69,33 @@ class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, Searchable
...
         CLog("Search for:", term)
         let searchTerm = cleanup(term: term)
+        let arabicSearchTerm = searchTerm.replacingOccurrences(of: regex,
+                                                               with: "_",
+                                                               options: .regularExpression)
+        let pattern = generateRegexPattern(term: term)
         let query = Database.table
-            .select(Database.Columns.snippet, Database.Columns.sura, Database.Columns.ayah)
-            .filter(Database.Columns.text.match("\(searchTerm)"))
-            .limit(300)
```
ahmedre: note - i intentionally dropped the `limit` because this opens up the search a lot wider, even though a lot fewer results come in.

Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -39,22 +69,33 @@ class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, Searchable
...
+            .select(Database.Columns.text, Database.Columns.sura, Database.Columns.ayah)
+            .filter(Database.Columns.text.like("%\(arabicSearchTerm)%"))
```
ahmedre: this no longer uses `match` because it takes advantage of the `_` character in `like`, which matches any character. essentially, we replace any of the characters that could have valid other options (ا for example) with an `_` to match anything. the postprocessing step them ensures that whatever filled that `_` is what we want.

sadly, that also means we have to highlight ourselves.

Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -119,6 +160,29 @@ class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, Searchable
...
+            if let replacement = replacements[char] {
+                regex = "\(regex)[\(replacement)]"
+            } else {
+                regex = "\(regex)\(char)"
+            }
+        }
+        return regex + ")"
+    }
+
+    private func highlightResult(text: String, regexPattern: String) -> String {
```
ahmedre: this highlights occurrences of the text within the result

Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -119,6 +160,29 @@ class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, Searchable
         }
         return cleanedTerm.lowercased()
     }
+
+    private func generateRegexPattern(term: String) -> String {
```
ahmedre: this generates the regex we run on each result from the database - it's basically the search term, with any of the possible characters that have alternatives (the keys of the `replacements` table) replaced with an optional set of any of their replacements.

ahmedre:
some notes:
- highlighting of results is partially broken (though it is broken today as well - ex in today's app, do a search for `قمر` and look at the result - you'll see the shaping broke between the `ال` and `قمر`).
- i found an existing bug that i didn't know how to fix - but it's unrelated to this patch (since i could repro without it) but if you search for `قمر`, the first result (the sura name result) has broken highlighting.

Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -39,22 +69,33 @@ class SQLiteSearchableAyahTextPersistence: ReadonlySQLitePersistence, Searchable
...
-    private func rowsToResults(_ rows: AnySequence<Row>, term: String) throws -> [SearchResult] {
+    private func rowsToResults(_ rows: AnySequence<Row>,
+                               term: String,
+                               regexPattern: String) throws -> [SearchResult] {
         var results: [SearchResult] = []
         for row in rows {
-            let text = row[Database.Columns.snippet]
+            let text = row[Database.Columns.text]
+            if text.range(of: regexPattern, options: .regularExpression) == nil {
+                continue
```
mohamede1945: Sorry, I didn't get this one. Why we are skipping results that doesn't match the regex? What happens if the user searches for `محمد` for example?

Quran/SQLiteSearchableAyahTextPersistence.swift
mohamede1945: What happens if the user includes any of the reserved regex characters? I think the search term is cleaned up before getting to this point, right?

Quran/SQLiteSearchableAyahTextPersistence.swift
```diff
@@ -158,4 +222,18 @@ private extension String {
         }
         return ranges.reversed()
     }
+
+    func regexRanges(of term: String) -> [Range<String.Index>] {
+        var ranges: [Range<String.Index>] = []
+        var maximum: Range<String.Index>?
+        while true {
```
mohamede1945: Can we use `NSRegularExpression.matches` from [here](https://developer.apple.com/documentation/foundation/nsregularexpression/1412446-matches) instead? We will remove the need for the `while true`.

mohamede1945:
Overall looks good to me. Have few minor questions, but nothing major.

PR #328, opened 2020-02-05
fix #211: shows image next to translation

### LINKED ISSUE #211: A Better Way to Show Word-By-Word Translation

aboualiaa:
This is an attempt at fixing #211. The code shows an image of the word next to the text in the popover.

This solution needs at least iOS 10.0 to work.

Some Issues that I couldn't solve:
* The image that gets displayed is somewhat small, due to limitations in the Popover pod
* It quite a few calculations every time the function is called, not resource friendly
* Not tested on a real device

[Video](https://youtu.be/39pVtjO1Xpk)


mohamede1945:
Wow! The video looks awesome! 

Jazak Allah khyran for implementing such a great UX improvement!

PR #363, opened 2022-01-15
Fail CI if any xcodebuild commands fail

mohamede1945:


mohamede1945:
Merging it since the test case fails already, this PR makes the CI fails if any test case fails.

PR #398, opened 2022-11-13
Implement logic to check for whether screen size can support 2 pages

zukhan:


mohamede1945:
Jazak allah khyran, Zubair!

PR #400, opened 2022-12-28
'Play' subtext not correct when multiple ayahs selected

zukhan:
**Before:**
before

**After:**
after


Sources/Localization/Resources/ar.lproj/Localizable.strings
```diff
@@ -118,6 +118,7 @@
 "ayah.menu.play-end-juz" = "إلى نهاية الجزء";
 "ayah.menu.play-end-surah" = "إلى نهاية السورة";
 "ayah.menu.play-end-page" = "إلى نهاية الصفحة";
+"ayah.menu.play-selected-verses" = "الآيات المختارة";
```
mohamede1945, line 121: I changed this from آيات مختارة to الآيات المختارة.

mohamede1945 (approved):
Amazing work, mashaa'Allah!

PR #406, opened 2023-01-19
WIP: Do not include basmallah before first verse when ayah selection is repeating

zukhan:
When a user has selected ayahs for repeating, we should exclude the basmallah from playing before the first verse of a surah. A common use case for repeating ayahs is to memorize and by excluding the basmallah, the user can focus on just the ayahs they are trying to memorize.


ahmedre:
@zukhan I realize you closed this, but fwiw, what we do on Quran for Android is gate this depending on repetition or not - i.e. if you're playing the first ayah of any sura (except Fatiha and Tawbah), we play the basmalah, unless it's a repeat of the ayah, in which case we don't.

PR #407, opened 2023-01-19
Add ayah menu item to allow playing selected verses in a loop

zukhan:
### Summary:

This change allows the user the ability to easily repeat selected verse(s). Currently: 

1. If you want to repeat multiple selected verses, you have to select the verses, click on "Play selected ayahs", navigate to audio options on the bottom right, click on "loop" in "Play set of verses" and then click on the Play button. 

2. If you want to repeat only one selected verse, you have to select multiple verses first, click on "Play selected ayahs", navigate to audio options on the bottom right, change the end verse to be the same as the start verse, then click on "loop" in "Play set of verses", then click on the Play button.

This change allows repeating one or more verses with a single menu item press after selecting the verse(s).

### Screenshots

**Single Verse Selected**

single-verse-selected

**Multiple Verses Selected**

multiple-verses-selected

### DIFF
--- a/Sources/Localization/Resources/en.lproj/Localizable.strings
+++ b/Sources/Localization/Resources/en.lproj/Localizable.strings
@@ -121,7 +121,9 @@
 "ayah.menu.play-end-juz" = "to the end of Juz'";
 "ayah.menu.play-end-surah" = "to the end of Surah";
 "ayah.menu.play-end-page" = "to the end of Page";
-"ayah.menu.play-selected-verses" = "selected verses";
+"ayah.menu.repeat" = "Repeat";
+"ayah.menu.selected-verse" = "selected verse";
+"ayah.menu.selected-verses" = "selected verses";
 "ayah.menu.highlight" = "Highlight";
 "ayah.menu.highlight-select-color" = "select color";
 "ayah.menu.add-note" = "Add Note";
--- a/Sources/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Sources/Localization/Resources/ar.lproj/Localizable.strings
@@ -120,7 +120,9 @@
 "ayah.menu.play-end-juz" = "إلى نهاية الجزء";
 "ayah.menu.play-end-surah" = "إلى نهاية السورة";
 "ayah.menu.play-end-page" = "إلى نهاية الصفحة";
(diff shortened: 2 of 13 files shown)


Sources/Localization/Resources/ar.lproj/Android.strings
```diff
@@ -181,6 +181,7 @@
 "prefs_use_arabic_summary_on" = "سيتم تحويل التطبيق إلى اللغة العربية";
 "prefs_use_arabic_title" = "الوضع العربي";
 "play" = "تشغيل";
+"repeat" = "كرر";
```
mohamede1945, line 184: The Android.strings files are autogenerated from the Android repo, let's not use it. Let's use Localizable.strings for strings dedicated to iOS only, please.

Sources/Localization/Resources/en.lproj/Localizable.strings
```diff
@@ -121,7 +121,9 @@
 "ayah.menu.play-end-juz" = "to the end of Juz'";
 "ayah.menu.play-end-surah" = "to the end of Surah";
 "ayah.menu.play-end-page" = "to the end of Page";
-"ayah.menu.play-selected-verses" = "selected verses";
+"ayah.menu.play-selected-verses-once" = "selected verses (once)";
```
mohamede1945, line 124: I don't think "Once" adds too much value. It also increases the width of the Ayah menu, not a big deal, but something to note.

mohamede1945 (changes requested):
Jazak Allah khyran for the great contribution!

Sources/Localization/Resources/ar.lproj/Android.strings
zukhan: Good to know, thanks!

zukhan: Fixed.

Sources/Localization/Resources/en.lproj/Localizable.strings
zukhan: Removed it.

mohamede1945 (approved):

mohamede1945:
Jazak Allah khyran, Zubair!

PR #416, opened 2023-03-31
Update Readme with a more accurate diacritic mark

Abid-dev:
Update Readme with a more accurate diacritic mark for the Arabic letter hamza


mohamede1945:
Thank you!

PR #482, opened 2023-06-04
Remove test targets from build action

mohamede1945:


mohamede1945:
Thought it would reduce build time, but it didn't.

PR #558, opened 2023-08-17
Fix russian transliteration strings for surahs titles

aziev:
Titles for some surahs are incorrect in russian. This PR fixes them.


mohamede1945:
Thank you for your contribution. However, I'm sourcing the translations from the Android repository. It would be best if you could make the fix there. Before the upcoming release, I'll import the updated translations.

Here's the file for your reference: [sura_names.xml in the Android repo](https://github.com/quran/quran_android/blob/0802bc019a4360f82d61be43df99183fc581bc51/app/src/main/res/values-ru/sura_names.xml).

aziev:
@mohamede1945 it seems to be fixed in the android app.
جزاك الله خيرا

mohamede1945:
Awesome! Jazak Allah Khairan for your help and verification, @aziev! I'll go ahead and merge this PR. That way, the next time I import translations, I can use your changes as a reference to ensure everything matches up. Thank 

you!

Issue #569, opened 2023-10-20
Please resolve compilation issues for Xcode 15 and create a version release instead of us having to point to main branch.


takasurazeem:
I am using Quran-iOS in a project, and it fails to compile on Xcode-Cloud due to following error:
an out-of-date resolved file was detected at /Volumes/workspace/repository/Al-Project.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved, which is not allowed when automatic dependency resolution is disabled; please make sure to update the file to reflect the changes in dependencies. Running resolver because the following dependencies were added: 'genericdatasource' (https://github.com/GenericDataSource/GenericDataSource)fatalError2023-10-19 18:17:42.184 xcodebuild[3497:15169] Writing error result bundle to /var/folders/8h/b7vb_p_j6sq0j82g8ts1jjv00000gn/T/ResultBundle_2023-19-10_18-17-0042.xcresult

PR #570, opened 2023-10-20
Update package.swift file

takasurazeem:
Consistent package names avoid any SPM errors in CI/CD pipelines. https://stackoverflow.com/a/63079910/3077444


takasurazeem:
This does not address https://github.com/quran/quran-ios/issues/569. Please fix the project for Xcode 15 soon and create a release so that we no longer have to point to the main branch. جزاك الله خیراً

mohamede1945 (approved):

PR #588, opened 2023-11-20
Translate missing localization using ChatGPT 4.5

mohamede1945:
Generated ChatGPT translations file: [chatgpt-translations.json](https://github.com/quran/quran-ios/files/13405490/chatgpt-translations.json)

### DIFF
--- a/Core/Localization/Resources/de.lproj/Localizable.strings
+++ b/Core/Localization/Resources/de.lproj/Localizable.strings
@@ -2,72 +2,276 @@
 
 // MARK: - General
 
+// Metadata: Translated by ChatGPT; human review required.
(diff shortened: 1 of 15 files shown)


ahmedre:
Wow Masha'Allah someone got early access to gpt 4.5? 😂

mohamede1945:
LOL ... I recall seeing 4.5 somewhere, but now I can't find it 😂 

Unfortunately, handling the translations wasn't easy due to their volume, and I got blocked twice for reaching the limit. Moving forward, it should be simpler as I'll only need to generate translations for new or updated keys, rather than all of them.

ahmedre:
Using their api would help here but would cost per input/output token.

PR #591, opened 2023-11-25
Remove Duaa to Gaza since Apple doesn't like it

mohamede1945:
Apple has rejected the App updates 6 times due to a Duaa to Gaza. They didn't explicitly mention it in their rejection, but they give me a nonsensical reason about a dispute that I won in 2017 (6 years ago). 

---
**May Allah lift the oppression on our brothers and sisters in Gaza.**


mohamede1945:
The update was approved after I removed the Duaa, which is disappointing. Although Apple did not explicitly state it, the pattern of rejections for the same nonsense reason, approving only after the Duaa's removal, strongly suggests that it was the cause of the issue.

Issue #648, opened 2024-08-20
Exporting Highlights


yismailuofa:
Hi there, is there a way to export your highlights/notes in some text based format ?

mohamede1945:
No, but open to contribution.

yismailuofa:
I got a basic version working on this draft PR here: https://github.com/quran/quran-ios/pull/649.

I haven't worked with Swift too much so the code probably isn't the best, but if you could take a look that would be great.

mohamede1945:
You're kidding Youssef, right? The code looks amazing Mashaa'Allah. Also, glad you found out the example project.

yismailuofa:
Hello again, is it possible to get this change released into the main iOS app?

mohamede1945:
I've just submitted it for review.

![image](https://github.com/user-attachments/assets/9b759d44-ee91-45f5-9742-f1eafb2d74c6)

PR #649, opened 2024-08-20
Add Sharing to Notes

yismailuofa:
Adds a basic sharing feature to the Notes page.

https://github.com/user-attachments/assets/be99339a-08c8-4108-9cd7-19d790977dd3

### DIFF
--- a/Features/NotesFeature/NotesViewController.swift
+++ b/Features/NotesFeature/NotesViewController.swift
@@ -18,7 +18,7 @@ final class NotesViewController: UIHostingController<NotesView> {
 
         initialize()
     }
-    
+
     @available(*, unavailable)
     @MainActor
     dynamic required init?(coder aDecoder: NSCoder) {
@@ -53,16 +53,32 @@ final class NotesViewController: UIHostingController<NotesView> {
                     image: UIImage(systemName: "square.and.arrow.up"),
                     style: .plain,
                     target: self,
-                    action: #selector(shareAllNotes))])
-    }    
+                    action: #selector(shareAllNotes)
+                ),
+            ]
+        )
+    }
 
-    @objc private func shareAllNotes() {
+    @objc
+    private func shareAllNotes() {
         Task {
-            let notesText = await viewModel.prepareNotesForSharing()
-            let activityViewController = UIActivityViewController(activityItems: [notesText], applicationActivities: nil)
-            activityViewController.popoverPresentationController?.sourceView = self.view
-            
-            present(activityViewController, animated: true, completion: nil)
+            do {
+                let notesText = try await viewModel.prepareNotesForSharing()
+
+                let activityViewController = UIActivityViewController(activityItems: [notesText], applicationActivities: nil)
+
+                // iPad support
+                let view = navigationController?.view
+                let viewBound = view.map { CGRect(x: $0.bounds.midX, y: $0.bounds.midY, width: 0, height: 0) }
+                activityViewController.modalPresentationStyle = .formSheet
+                activityViewController.popoverPresentationController?.permittedArrowDirections = []
+                activityViewController.popoverPresentationController?.sourceView = view
+                activityViewController.popoverPresentationController?.sourceRect = viewBound ?? .zero
+
+                present(activityViewController, animated: true)
+            } catch {
+                showErrorAlert(error: error)
+            }
         }
     }
 }
--- a/Features/NotesFeature/NotesViewController.swift
+++ b/Features/NotesFeature/NotesViewController.swift
@@ -18,7 +18,7 @@ final class NotesViewController: UIHostingController<NotesView> {
 
         initialize()
     }
-
+    
     @available(*, unavailable)
     @MainActor
     dynamic required init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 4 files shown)


Features/NotesFeature/NotesViewController.swift
```diff
@@ -47,7 +47,22 @@ final class NotesViewController: UIHostingController<NotesView> {
...
+                    style: .plain,
+                    target: self,
+                    action: #selector(shareAllNotes))])
+    }    
+
+    @objc private func shareAllNotes() {
+        Task {
+            let notesText = await viewModel.prepareNotesForSharing()
+            let activityViewController = UIActivityViewController(activityItems: [notesText], applicationActivities: nil)
+            activityViewController.popoverPresentationController?.sourceView = self.view
```
mohamede1945, line 63: I think this might crash on iPads or the popover arrow might not be pointing to the share button. Could you please verify that?

Features/NotesFeature/NotesViewModel.swift
```diff
@@ -66,10 +71,40 @@ final class NotesViewModel: ObservableObject {
...
+                                    
+                                    
+                    let verseText = verses.joined(separator: "\n")
+                    
+                    notesText += "\(noteText)\(verseText)\n\n"
+                }
+                
+                return notesText
+            }
+        } catch {
```
mohamede1945, line 98: I would rather not catch the error and show an error alert to the user. Use: https://github.com/quran/quran-ios/blob/4df3eb4a8c91571c18d9e8e38592bf7a0bad6b50/UI/NoorUI/BaseControllers/UIViewController%2BError.swift#L26

Features/NotesFeature/NotesViewModel.swift
```diff
@@ -66,10 +71,40 @@ final class NotesViewModel: ObservableObject {
         }
     }
 
+    func prepareNotesForSharing() async -> String {
+        do {
+            return try await crasher.recordError("Failed to share notes") {
+                var notesText = ""
+                for note in await notes {
+                    let noteText = if let noteContent = note.note.note, noteContent != "" {
+                        "Note: \(noteContent.trimmingCharacters(in: .newlines))\n"
```
mohamede1945, line 80: This needs to be localized. Maybe we can just not have "Note: " as a prefix and just show the text directly.

Features/NotesFeature/NotesViewModel.swift
```diff
@@ -66,10 +71,40 @@ final class NotesViewModel: ObservableObject {
...
+                    } else {
+                        ""
+                    }
+                    
+                    let verses = try await textRetriever.textForVerses(Array(note.note.verses))
+                        .filter { !$0.isEmpty }
+                        .map { $0.hasSuffix("﴾") ? String($0.dropLast()) : $0 }
```
mohamede1945, line 87: Do you mind explaining this logic? Just want to see if there's a problem in the textForVerses method.

mohamede1945 (changes requested):
Mashaa'Allah! Amazing implementation. Jazak allah khayrn.

I've added a few minor comments, but overall looks great.

mohamede1945:
For the SwifFormat, please run `make format-autocorrect` and it should fix any formatting issues.

Features/NotesFeature/NotesViewController.swift
yismailuofa: You were right, added some code from [here](https://github.com/quran/quran-ios/blob/4df3eb4a8c91571c18d9e8e38592bf7a0bad6b50/Features/SettingsFeature/UIViewController%2BShare.swift#L19-L24) to fix it.
simulator_screenshot_E0926277-A1E8-42F2-865F-0533B949504F

Features/NotesFeature/NotesViewModel.swift
yismailuofa: The arabic text from `textForVerses` has a trailing bracket that looks odd so I opted to drop it from the string.

It comes from [here](https://github.com/quran/quran-ios/blob/4df3eb4a8c91571c18d9e8e38592bf7a0bad6b50/Domain/QuranTextKit/Sources/ShareableText/ShareableVerseTextRetriever.swift#L66-L69) I believe. You can also see some more examples of the output [here](https://github.com/quran/quran-ios/blob/4df3eb4a8c91571c18d9e8e38592bf7a0bad6b50/Domain/QuranTextKit/Tests/ShareableVerseTextRetrieverTests.swift#L59-L63).

I'm not sure if it's a bug or a result of some other logic since the opening bracket at the end matches the closing bracket at the next ayah.

yismailuofa: image

mohamede1945: Oh, I see. This is actually an issue with renderer when it comes to combining Arabic and English letters. The brackets `()` are part of English (left-to-right) while the text is arabic (right-to-left) and end of the ayah is an arabic number surrounded by the brackets (e.g. `(1)`). So most renderer just show them broken and show the closing bracket at the start of the sentence from right. So what we could do instead is to replace `(1)`, `(2)`, ... these with just some other separator that would render right (e.g. `۞`). What do you think?

mohamede1945: I'm going to approve the PR, but feel free to fix this as a follow up PR or I can fix it myself this weekend.

mohamede1945 (approved):

yismailuofa:
closes #648

Features/NotesFeature/NotesViewModel.swift
yismailuofa: Thanks! I opened up a PR with a bit of a hacky solution but it might work. It's also not tested on anything besides English & Arabic, but maybe this could be a start.

https://github.com/quran/quran-ios/pull/650

Issue #348, opened 2021-05-08
Allow other apps to launch this app and jump to a particular Ayah


adil-hussain-84:
I'm building a couple of Hifdh revision assistance apps for iOS which will offer the ability to jump the user to see a particular Ayah in the [quran.com](https://quran.com) web app. I will do this by opening a URL of the form `https://quran.com/\(surahNumber)/\(ayahNumber)`.

Given that this app is owned by [quran.com](https://quran.com), it would be great if this app could register itself as capable of opening [https://quran.com](https://quran.com) links as described in the [Supporting Universal Links in Your App](https://developer.apple.com/documentation/xcode/supporting-universal-links-in-your-app) Apple Developer guide. A user of my apps will then be navigated into this app rather than the web app when this app is installed on the user's device.

In addition to supporting [quran.com](https://quran.com) universal links, it may be a good idea to support the `quran` custom URL scheme so links of the form `quran://\(surahNumber)/\(ayahNumber)` are also accepted as described in the [Defining a Custom URL Scheme for Your App](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app) Apple Developer guide. This is the type of URL accepted by the [Quran for Android](https://play.google.com/store/apps/details?id=com.quran.labs.androidquran) and [Al Quran (Tafsir & by Word)](https://play.google.com/store/apps/details?id=com.greentech.quran) Play Store apps.

mohamede1945:
Yeah, totally agree. Thanks for the suggestion.

wedkemail:
I haven't done this before, but i don't mind giving it a try! will create a branch to try it

iymorshe:
Any update on this? Looking for a URL scheme for Quran app.

mohamede1945:
No not yet. But contribution is more than welcome.

iman2153:
This brother's PR looks like a start
https://github.com/quran/quran-ios/pull/657

PR #657, opened 2024-09-06
Custom URL Scheme Implementation

hassan31:
Implemented custom url scheme to support launching the Quran Application from outside the application
For example, enter this in browser `quran-ios://` and it will redirect you to the Quran Application

Example Video

https://github.com/user-attachments/assets/cafcc9f4-5c24-4ebd-b6fb-1b680ab69f7c

### DIFF
--- a/Example/QuranEngineApp/Classes/AppDelegate.swift
+++ b/Example/QuranEngineApp/Classes/AppDelegate.swift
@@ -49,27 +49,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let downloadManager = container.downloadManager
         downloadManager.setBackgroundSessionCompletion(completionHandler)
     }
-    
-    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
-        // Check if the URL scheme matches the custom scheme
-        if url.scheme == "quran-ios" {
-            // Parse the URL to determine the desired action
-            let path = url.host ?? ""
-            
-            // Implement custom logic to navigate to a specific view or perform an action
-            // Example: Navigate to a specific surah or ayah
-            handleCustomURL(path: path)
-            
-            return true
-        }
-        return false
-    }
-
-    private func handleCustomURL(path: String) {
-        // Example logic to handle the custom URL path
-        // This could involve navigating to a specific part of the app
-        // For instance, opening a specific Surah or Ayah
-    }
 
     // MARK: Private
 
--- a/Example/QuranEngineApp/Classes/AppDelegate.swift
+++ b/Example/QuranEngineApp/Classes/AppDelegate.swift
@@ -49,6 +49,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let downloadManager = container.downloadManager
         downloadManager.setBackgroundSessionCompletion(completionHandler)
     }
+    
+    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+        // Check if the URL scheme matches the custom scheme
+        if url.scheme == "quran-ios" {
+            // Parse the URL to determine the desired action
+            let path = url.host ?? ""
+            
+            // Implement custom logic to navigate to a specific view or perform an action
+            // Example: Navigate to a specific surah or ayah
+            handleCustomURL(path: path)
+            
+            return true
+        }
+        return false
+    }
+
+    private func handleCustomURL(path: String) {
+        // Example logic to handle the custom URL path
+        // This could involve navigating to a specific part of the app
+        // For instance, opening a specific Surah or Ayah
+    }
 
     // MARK: Private
 
--- a/Example/QuranEngineApp/Info.plist
+++ b/Example/QuranEngineApp/Info.plist
@@ -2,6 +2,17 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLName</key>
+			<string>com.quran.QuranEngineApp</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>quran-ios</string>
+			</array>
+		</dict>
+	</array>
 	<key>UIApplicationSceneManifest</key>
 	<dict>
 		<key>UIApplicationSupportsMultipleScenes</key>
--- a/Example/QuranEngineApp/Info.plist
+++ b/Example/QuranEngineApp/Info.plist
@@ -2,17 +2,18 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
-	<key>CFBundleURLTypes</key>
-	<array>
-		<dict>
-			<key>CFBundleURLName</key>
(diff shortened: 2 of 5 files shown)


iman2153 (approved):
lgtm

hassan31:
@mohamede1945 could you please review this pr and approve so that it can be merged. Thanks

mohamede1945:
Salam alikom Muhammad, thank you for the contribution. But the code is empty. There's nothing implemented to parse the query parameters and navigate to a specific Ayah as mentioned in #348.

hassan31:
@mohamede1945 Wa Alaikumuslaam, Yes, this is the partial implementation, in which I have launched the application from outside link. In followup pr I will write the code to parse. So if you could approve and merge that would be really great.

Example/QuranEngineApp/Classes/AppDelegate.swift
```diff
@@ -49,6 +49,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let downloadManager = container.downloadManager
         downloadManager.setBackgroundSessionCompletion(completionHandler)
     }
+    
+    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+        // Check if the URL scheme matches the custom scheme
+        if url.scheme == "quran-ios" {
```
mohamede1945, line 55: Let's handle "quran" and "quran-ios" schemes please.

Example/QuranEngineApp/Classes/AppDelegate.swift
```diff
@@ -49,6 +49,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let downloadManager = container.downloadManager
         downloadManager.setBackgroundSessionCompletion(completionHandler)
     }
+    
+    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+        // Check if the URL scheme matches the custom scheme
```
mohamede1945, line 54: Please remove this redundant comment. The code is clear. Same with the other inline comments below.

Example/QuranEngineApp/Classes/AppDelegate.swift
```diff
@@ -49,6 +49,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        if url.scheme == "quran-ios" {
+            // Parse the URL to determine the desired action
+            let path = url.host ?? ""
```
mohamede1945, line 57: host is not the path. We should look at the url.path rather than the host.

Example/QuranEngineApp/Classes/AppDelegate.swift
```diff
@@ -49,6 +49,27 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            // Implement custom logic to navigate to a specific view or perform an action
+            // Example: Navigate to a specific surah or ayah
+            handleCustomURL(path: path)
+            
+            return true
+        }
+        return false
+    }
+
+    private func handleCustomURL(path: String) {
```
mohamede1945, line 68: `handle` seems very generic. I would rename it to `navigateTo(path: String)`. Also let's make this method return false if it can't parse and navigate to the path.

Example/QuranEngineApp/Info.plist
```diff
@@ -2,6 +2,17 @@
...
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
+	<array>
+		<dict>
+			<key>CFBundleURLName</key>
+			<string>com.quran.QuranEngineApp</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>quran-ios</string>
```
mohamede1945, line 12: Let's add an entry for `quran` scheme please.

mohamede1945 (changes requested):
Sounds good. I've added a couple of comments, please take a look. Also, please make sure to run `make format-autocorrect` to fix formatting issues.

Thank you!

mohamede1945:
Oh actually, please move the validation, parsing and navigation code to the QuranEngine instead of the example app. The example app would simply just implement the delegate method but then calls the QuranEngine to do the actual logic, this way the code can be shared among different apps.

Example/QuranEngineApp/Info.plist
hassan31: addressed

Example/QuranEngineApp/Classes/AppDelegate.swift
hassan31: addressed

Example/QuranEngineApp/Classes/AppDelegate.swift
hassan31: addressed

Example/QuranEngineApp/Classes/AppDelegate.swift
hassan31: addressed

Example/QuranEngineApp/Classes/AppDelegate.swift
hassan31: addressed

hassan31:
> Oh actually, please move the validation, parsing and navigation code to the QuranEngine instead of the example app. The example app would simply just implement the delegate method but then calls the QuranEngine to do the actual logic, this way the code can be shared among different apps.

AOA @mohamede1945 I have addressed your comments, kindly review again. Also regarding move the validation, parsing and navigation code to QuranEngine, i can see the same code there, but in pr it reflects in Example/QuranEngine app. May be if you guide how to move then I would do. Please check below screenshot.

Screenshot 2024-09-08 at 12 16 07 AM

mohamede1945:
You see [here](https://github.com/quran/quran-ios/blob/main/Example/QuranEngineApp/Classes/AppDelegate.swift#L26-L29) how the code in the example project delegates the business logic to the DownloadManager and ReadingResourcesService. These live in the [QuranEngine library](https://github.com/quran/quran-ios/blob/main/Package.swift) not in the Example project. The example project just delegates the call to these services.

In url scheme navigation logic, I think it makes sense to live in the AppStructureFeature.

hassan31:
> You see [here](https://github.com/quran/quran-ios/blob/main/Example/QuranEngineApp/Classes/AppDelegate.swift#L26-L29) how the code in the example project delegates the business logic to the DownloadManager and ReadingResourcesService. These live in the [QuranEngine library](https://github.com/quran/quran-ios/blob/main/Package.swift) not in the Example project. The example project just delegates the call to these services.
> 
> In url scheme navigation logic, I think it makes sense to live in the AppStructureFeature.

@mohamede1945 please review now, is this what you are expecting to move the logic to QuranEngine?

hassan31:
> You see [here](https://github.com/quran/quran-ios/blob/main/Example/QuranEngineApp/Classes/AppDelegate.swift#L26-L29) how the code in the example project delegates the business logic to the DownloadManager and ReadingResourcesService. These live in the [QuranEngine library](https://github.com/quran/quran-ios/blob/main/Package.swift) not in the Example project. The example project just delegates the call to these services.
> 
> In url scheme navigation logic, I think it makes sense to live in the AppStructureFeature.

@mohamede1945 can we connect through some communication tools, like slack or google meet, I wanted to learn more about the structure?

hassan31:
@mohamede1945 are you still available to review the latest changes?

mohamede1945:
Jazak Allah khyrn for the changes. I'm going to merge it after CI finishes. You can reach out to me on discord with the same username used here in github.

hassan31:
> Jazak Allah khyrn for the changes. I'm going to merge it after CI finishes. You can reach out to me on discord with the same username used here in github.

Merge is blocked, because of some checks are not successful. I also installed the swiftlint, but looks good on my side. May be you can help to find what is wrong with which format.

Also, I sent you request on discord. Please accept. Jazakallah Khair.

mohamede1945 (approved):

PR #659, opened 2024-09-13
Custom URL Scheme Implementation Supporting Sura and Verse Number

iman2153:
Implemented custom url scheme to support launching the Quran Application from outside the application
For example, enter this in browser quran-ios:// and it will redirect you to the Quran Application

### DIFF
--- a/Example/QuranEngineApp/Classes/AppDelegate.swift
+++ b/Example/QuranEngineApp/Classes/AppDelegate.swift
@@ -49,27 +49,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         let downloadManager = container.downloadManager
         downloadManager.setBackgroundSessionCompletion(completionHandler)
     }
-    
-    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
-        // Check if the URL scheme matches the custom scheme
-        if url.scheme == "quran-ios" {
-            // Parse the URL to determine the desired action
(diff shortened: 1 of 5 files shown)


iman2153:
Asalamuwalaikum @mohamede1945 want some thoughts on implementing the full #348 as seen in #657.

I need to implement error handling:
1.   suras above 114 
2. out of range verses based on each sura

Do you want the parameters from 0 to 113 or 1 to 114? I believe the first makes more sense as its the default in the project atm.

Any PRs I can reference for linking to certain views in the App?

mohamede1945:
**Jazak Allah Khairan** for your contribution! I truly appreciate the effort you put into this. I have a few suggestions to help improve the implementation before it can be merged:

1. The current implementation is a great start, but it's a bit incomplete. While it handles parameter parsing well, it still needs the actual navigation logic to be fully functional.

2. When you get a chance, please rebase your work on top of the latest main branch to ensure compatibility and allow me to only review updated code in the PR.

3. To make the code more robust, consider adding unit tests for the business logic (such as parameter parsing). This isn’t required for UI and navigation code, but it helps ensure the core functionality works as expected.

Regarding your questions:
1. We should use `AyahNumber`, and it supports returning `nil` for invalid ayahs.
2. `AyahNumber` works with values from 1 to 114.
3. Additionally, it would be helpful to include navigation to a surah, not just an ayah.

Again, thank you for your contribution! I look forward to seeing your updates. Please let me know if you have any questions or need further guidance.

PR #667, opened 2024-09-17
Fix Arabic Number Bug and Improve Notes Sharing

### LINKED ISSUE #666: Tafseer view and notes view has English numbers for Ayat despite app language being Arabic
| Tafseer | Notes  |
|--------|--------|
|  |  |

yismailuofa:
1.  Fixes #666, 
I think maybe Apple changed the locale keys in the new update ? The docs say it uses [BCP-47](https://developer.apple.com/documentation/foundation/locale/2293563-init). So updated the string to use [this](https://www.techonthenet.com/js/language_tags.php).

2. Refactors the sharing abit to make it more consistent, since sometimes English and Arabic would be on the same line, now we add newlines properly I think.

3. Updates the ayah sharing to use the same logic as [copy](https://github.com/quran/quran-ios/blob/c40fb5c958ff0f572e085e1b1ec67c716b68ca78/Features/AyahMenuFeature/AyahMenuViewModel.swift#L136-L145). Before copy would have newlines and sharing didn't but now they should both have it.

<div style="display: flex; gap: 10px;">
    Image 7630
    Image 7631
</div>

### DIFF
--- a/Core/Localization/NumberFormatter+Extension.swift
+++ b/Core/Localization/NumberFormatter+Extension.swift
@@ -49,7 +49,7 @@ extension NumberFormatter {
 
     public static var arabicNumberFormatter: NumberFormatter = {
         let formatter = NumberFormatter()
-        formatter.locale = Locale(identifier: "ar")
+        formatter.locale = Locale(identifier: "ar-SA")
         return formatter
     }()
 }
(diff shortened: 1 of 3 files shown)


mohamede1945 (approved):

mohamede1945:
Jazak Allah Khyrn Youssef for fixing them. 

I tried to repro #666 but I couldn't.

yismailuofa:
@mohamede1945 For reference, I’m on ios18. Maybe #666 is as well?

mohamede1945:
Oh, I see. I'm still on ios 17. Also, I've sent a new version for review.

PR #681, opened 2024-12-31
Authentication – Refresh, persistence & API

mohannad-hassan:
This PR should complete all the functionalities needed to handle the infrastructure of authentication, bar the logout API. This should be handled when adding the details of the login and profile UI. 

## Main Changes
- Handling refreshing, persistence and authenticating APIs
- Restructuring the client to ease testing the public APIs. 

## Some Design Points
- I've decided to maintain the configurations setup, at least for now. The public state of the client needs to indicate two non-usable states (not-configured and not-authenticate), so I saw that it's better to unify access through a non-optional type for both of them. 
- I expect that the we might need to run startup actions (eagerly, or lazily) related to data synchronization. Currently, one action is done to restore and refresh the authentication state. 
- This is one point where design might evolve noticeably. The two previous points are expected to be revisited.


Data/AuthenticationClient/Sources/Persistance.swift
```diff
@@ -0,0 +1,83 @@
...
+import Foundation
+import VLogging
+
+enum PersistanceError: Error {
+    case persistanceFailed
+    case retrievalFailed
+}
+
+/// An abstraction for secure persistance of the authentication state.
+protocol Persistance {
```
mohamede1945, line 17: I suggest creating a general `SecurePersistence` API (under Core/ directory) rather than one specifically for authentication.  

A recommended pattern can be seen in [this directory](https://github.com/quran/quran-ios/tree/3d34dadc1fb260e035b23649fd28f5b06a22f6b0/Core/Preferences), where `Preferences` is implemented as a service, and consumers define `PreferenceKey` objects to describe the data that needs to be persisted. I think we can do something similar with `SecurePersistence` and `SecurePersistenceKey`.

Also please add a couple of tests for this class.

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
```diff
@@ -0,0 +1,219 @@
...
+    let configuration = OAuthAppConfiguration(
+        clientID: "client-id",
+        redirectURL: URL(string: "callback://")!,
+        scopes: [],
+        authorizationIssuerURL: URL(string: "https://example.com")!
+    )
+
+    override func setUp() {
+        caller = OAuthCallerMock()
+        persistance = PersistanceMock()
```
mohamede1945, line 26: I recommend minimizing the use of mocks, as they can make tests more dependent on the mock behavior than the actual application code. Instead, I recommend focusing on mocking only the third-party API for authentication and avoid creating mocks for the code we own. This approach ensures tests remain closely aligned with the application's real behavior.

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    func testNoConfigurations() async throws {
+        XCTAssertEqual(sut.authenticationState, .notAvailable, "Expected to signal a not-configured state")
+        do {
+            try await sut.login(on: UIViewController())
+            XCTFail("Expected to throw an error if prompted to login without app configuration being set.")
+        } catch {
+            // Success
+        }
```
mohamede1945, line 37: Use AsyncAssertThrows

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
```diff
@@ -0,0 +1,219 @@
...
+
+        do {
+            try await sut.login(on: UIViewController())
+
+            XCTAssertTrue(persistance.clearCalled, "Expected to clear the persistance first")
+            XCTAssertEqual((persistance.currentState as? AutehenticationDataMock), state, "Expected to update the new state")
+            XCTAssertEqual(sut.authenticationState, .authenticated, "Expected the auth manager to be in authenticated state")
+        } catch {
+            XCTFail("Expected to login successfully -- \(error)")
+        }
```
mohamede1945, line 57: The do-catch is not needed. XCTest will fail the test if an except is thrown. Same in other tests below

mohamede1945 (changes requested):
Jazak Allah khair, Mohannad!  

I believe we need to refine the design of a few components to make them more reusable and minimize the chances of failure. Let me know if you have any questions or need clarification.  

Thank you!

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  OAuthClient.swift
+//  AuthentincationDataManager.swift
```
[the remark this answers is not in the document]
mohannad-hassan: Resorted to `client`. I see that nothing under Data is named "Service", and that it's a common name for the classes in domain.

Data/AuthenticationClient/Sources/AuthentincationDataManagerImpl.swift
```diff
@@ -0,0 +1,120 @@
...
+    private var cancellables = Set<AnyCancellable>()
+
+    private var appConfiguration: OAuthAppConfiguration?
+
+    private var state: AuthenticationData? {
+        didSet {
+            guard let state else { return }
+            state.stateChangedPublisher.sink { [weak self] _ in
+                self?.persist(state: state)
+            }.store(in: &cancellables)
```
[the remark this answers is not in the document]
mohannad-hassan: Good observation.

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
mohannad-hassan: I'm actually of the opinion that one should _avoid_ mocking 3rd party code. Not often that these would be open for subclassing or easy for subclassing. The class `OIDAuthState` of the library we're using proved to be so. It was open, but I'd have to mock four other types. Same issue prevented from stubbing it.  

Moreover, wrapping 3rd party code under abstractions allows to define the interactions between the main logic and the kind of responsibility we seek from the 3rd party code. In other words, we're adapting the 3rd party code to our needs - as much as the situation allows without awkward design.

mohannad-hassan: As for depending on mocks, I understand that this is a concern, however, I try to counter that by keeping mocks as small as possible, and relying only on providing ready responses (call succeeds or fails, call returns something, ...)  

I use the term 'mock' liberally, what I usually use are mostly stubs and spies. 

I expounded on that, because these are key techniques I use (dealing with 3rd party and testing doubles strategies.) We can discuss that face-to-face if there's something to be aligned.

Data/AuthenticationClient/Sources/AuthenticationData.swift
```diff
@@ -0,0 +1,121 @@
...
+    func getFreshTokens() async throws -> String {
+        fatalError()
+    }
+
+    override init() { }
+
+    required init(from decoder: any Decoder) throws {
+        fatalError()
+    }
+}
```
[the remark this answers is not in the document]
mohannad-hassan: This was supposed to be a protocol, but turned it into a class (similarly to Java's abstract classes.) The reason for this is the need for `Codable` conformance while also using the type `AuthenticationData` as an existential type. 

However, I did experiment with it now, and it seems it can be used as an existential type. I'll be updating this.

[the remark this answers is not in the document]
mohannad-hassan: BTW, I thought of extracting the OAuth wrappers (caller and data) out to a `Core` package, but these library-specific design discrepancies discouraged me from that.

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
```diff
@@ -32,13 +47,21 @@ public struct OAuthAppConfiguration {
...
     func set(appConfiguration: OAuthAppConfiguration)
 
     /// Performs the login flow to Quran.com
     ///
     /// - Parameter viewController: The view controller to be used as base for presenting the login flow.
     /// - Returns: Nothing is returned for now. The client may return the profile infromation in the future.
     func login(on viewController: UIViewController) async throws
+
+    /// Returns `true` if the client is authenticated.
+    func restoreState() async throws -> Bool
```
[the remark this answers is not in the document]
mohannad-hassan: Perhaps naming needs to be reconsidered. The purpose of this is to restore and reevaluate the status. The full picture of this API is to be used by sync logic. 

But your intuition is on point: this part is still not clear. The main remaining point is synchronization especially on startup, as this API is intended to be the base for that.

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
```diff
@@ -32,13 +47,21 @@ public struct OAuthAppConfiguration {
 
 /// Handles the OAuth flow to Quran.com
 ///
-/// Note that the connection relies on dicvoering the configuration from the issuer service.
-public protocol OAuthClient {
+/// Expected to be configuered with the host app's OAuth configuration before further operations are attempted.
+public protocol AuthentincationDataManager {
+    /// Sets the app configuration to be used for authentication.
     func set(appConfiguration: OAuthAppConfiguration)
```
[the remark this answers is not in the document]
mohannad-hassan: Okay then, I moved setting the configurations to the initialization. However, I'm still not comfortable to make it nullable. 

As explained in the PR's description: _The public state of the client needs to indicate two non-usable states (not-configured and not-authenticate), so I saw that it's better to unify access through a non-optional type for both of them._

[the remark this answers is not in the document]
mohamede1945: > The public state of the client needs to indicate two non-usable states (not-configured and not-authenticate)

Do you mind explaining why is that please?

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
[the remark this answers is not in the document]
mohamede1945: What I’ve observed in other auth libraries is that they typically provide functionality to check `isAuthenticated` and perform `login`, similar to your implementation. In addition to that they often include either an explicit `start` method or implicitly initiate the auth process during the initialization of the service/client class. So, that would be my recommendation.

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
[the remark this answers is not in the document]
mohannad-hassan: Services using the client need to guard against the state of not being authenticated. They also need to guard against the case of the client not having been configured (if cloned by someone.) These are the two non-usable states of the client. 

The client needs to expose the first state, so my first idea is to unify all states of the client (not-configured, non-authenticated & authenticated) through the `authenticationState` property. I favoured this over handling one state through nil checking and the other through a property.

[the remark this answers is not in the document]
mohannad-hassan: To be honest, a lot of the vagueness here will be cleared once we get to integrating the new APIs along the normal flow.

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
mohamede1945: Apologies, I’m not sure I fully understand your point. You mentioned wanting to avoid mocking but also referred to implementing small mocks. 

To clarify my perspective: tests should avoid using mocks/fakes/spies/stubs unless absolutely necessary. These cases include scenarios where the actual dependency cannot be used in tests (e.g., backend APIs or time-based APIs) or where using a mock provides a significant benefit (e.g., file system operations that could slow tests due to extensive reading and writing). 

If a mock is needed, it should be added to the following directory: [Core/SystemDependencies](https://github.com/quran/quran-ios/tree/main/Core/SystemDependencies). Additionally, a corresponding fake implementation should be placed in: [Core/SystemDependenciesFake](https://github.com/quran/quran-ios/tree/main/Core/SystemDependenciesFake). This approach ensures that we explicitly allowlist the APIs for which their actual implementation cannot be used.

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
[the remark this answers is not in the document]
mohamede1945: In general, a client instance should be configured for a consumer to use it. If the consumer does not configure the client, no client will be available for use. A null client serves as a clear indication that the app is not configured for authentication. This approach aligns with best practices followed by many system APIs, where the API is configured during initialization to ensure it can be used effectively.

There are additional nuances to consider, such as handling invalid configurations. Ideally, verifying the configuration during initialization would be great, but I assume this isn't feasible due to the need for backend API requests. Therefore, late verification of the configuration is acceptable. However, when early validation is possible, we should prioritize it and not defer it.

Data/AuthenticationClient/Sources/AuthentincationDataManager.swift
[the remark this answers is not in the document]
mohamede1945: Yeah Client works too. Thank you!

Data/AuthenticationClient/Sources/AuthenticationData.swift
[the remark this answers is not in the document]
mohamede1945: I believe using a protocol is a positive step forward. However, the current API and implementation handle multiple responsibilities, and I’d suggest breaking them up for better separation of concerns:

1. **Data Handling:** Managing the data, including serialization and deserialization.  
2. **State Change Listening:** Listening to `OIDAuthStateChangeDelegate` changes and notifying consumers.  
3. **Action Encapsulation:** Encapsulating actions that can be performed by `OIDAuthState`.

My suggestion is to separate the first responsibility from the second and third. For example:

1. Introduce an `AuthState` **struct** to handle serialization and deserialization of state details.  
2. Create an `AuthStateService` that listens for state changes and provides APIs such as `refreshToken` and `isAuthorized`. 

This approach ensures a cleaner design by dividing responsibilities across focused components.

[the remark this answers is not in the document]
mohamede1945: I also recognize that the library's API might be awkward at times, making it challenging to provide a more refined solution. However, I truly appreciate your patience and effort in overcoming these obstacles. 

Thank you so much for your hard work, Mohannad!

Data/AuthenticationClient/Tests/AuthentincationDataManagerTests.swift
mohannad-hassan: > You mentioned wanting to avoid mocking but also referred to implementing small mocks. 
I avoid mocking 3rd party code; in our case, that would be keychain access functions and the AppAuth-iOS library. My solution for this would be wrapping such code behind an abstraction and providing a mock that extends that. 

> tests should avoid using mocks/fakes/spies/stubs unless absolutely necessary 
I see. I largely agree. I usually mock things that represent complex logic and I'd think of mocking first for cross-package/cross-layer dependencies. But since this is the practice we're taking here, I'll adjust it. 

For our case though, we have:
- Keychain access, wrapped by `Persistence`. I can see this fits well with SystemDependencies. 
- The OAuth library, wrapped by `OAuthCaller`. Do you that sitting there?

mohannad-hassan:
@mohamede1945 
Here's an update: 
- The most noticeable update is the redesign of the OAuth client part, that is used by `AuthenticationClient`. The `OAuthService` is now responsible for executing actions and _updating state_. The new data, or `OAuthStateData` should be a simple struct now. 
- Extracted that part to a separate Core module. However, I'd like your input if anything is missing from the current mocks to be the kind of fakes we're using, before this is extracted out.
- Didn't extract `Persistence` out. It's now an internal type, so the API is fitted for that. I preferred not to put time into making it more general. Let me know what you think. 

There is one or two points that I've put off for now: 
- Defining an enum for the available scopes Quran.com has. 
- Currently, OAuth-flow errors revert the client to a non-authenticated state. It might be needed to differentiate server failures and networking issues and other classes. I think it'll be better, in terms of its UX, not to request the user to login again in the first case.

Core/OAuthService/OAuthService.swift
```diff
@@ -0,0 +1,45 @@
...
+///
+/// The service is assumed not to have any internal state. It's the responsibility of the client of this service
+/// to hold and persist the state data. Each call to the service returns an updated `OAuthStateData`
+/// that reflects the latest state.
+public protocol OAuthService {
+    func login(on viewController: UIViewController) async throws -> OAuthStateData
+
+    func getAccessToken(using data: OAuthStateData) async throws -> (String, OAuthStateData)
+
+    func refreshIfNeeded(data: OAuthStateData) async throws -> OAuthStateData
```
mohamede1945, line 36: nit: Can we rename refreshIfNeeded to refreshAccessTokenIfNeeded to make it clearer please.

Data/AuthenticationClient/Sources/AppAuthOAuthService.swift
```diff
@@ -0,0 +1,158 @@
...
+    init(appConfigurations: AuthenticationClientConfiguration) {
+        self.appConfigurations = appConfigurations
+    }
+
+    // MARK: Internal
+
+    func login(on viewController: UIViewController) async throws -> any OAuthStateData {
+        // Quran.com relies on dicovering the service configuration from the issuer,
+        // and not using a static configuration.
+        let serviceConfiguration = try await discoverConfiguration(forIssuer: appConfigurations.authorizationIssuerURL)
```
mohamede1945, line 51: Can we move this line to the AuthenticationClient and keep this implementation just a wrapper around 3p library.

Data/AuthenticationClient/Sources/Persistence.swift
```diff
@@ -0,0 +1,101 @@
...
+enum PersistenceError: Error {
+    case persistenceFailed
+    case retrievalFailed
+}
+
+/// An abstraction for secure persistence of the authentication state.
+protocol Persistence {
+    func persist(state: Data) throws
+
+    func retrieve() throws -> Data?
```
mohamede1945, line 20: To make this generic:

1. I think we need to make this keyed persistence to allow saving and retrieving different kind of information.
2. Can we move it under Core/ as a separate target please?

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
```diff
@@ -0,0 +1,302 @@
...
+        }
+        return try JSONEncoder().encode(data)
+    }
+
+    func decode(_ data: Data) throws -> any OAuthStateData {
+        try JSONDecoder().decode(AutehenticationDataMock.self, from: data)
+    }
+}
+
+private final class OAuthServiceMock: OAuthService {
```
mohamede1945, line 198: Let's move this to Core because we will need to use it in more tests as we integrate auth with other features.

Data/AuthenticationClient/Sources/AppAuthOAuthService.swift
```diff
@@ -0,0 +1,158 @@
+//
```
mohamede1945, line 1: Since this is an implementation of Core/OAuthService. Can we have it under maybe Core/OAuthServiceImpl target?

Data/AuthenticationClient/Sources/AuthenticationClient.swift
```diff
@@ -0,0 +1,65 @@
...
+/// Expected to be configuered with the host app's OAuth configuration before further operations are attempted.
+public protocol AuthenticationClient {
+    /// Performs the login flow to Quran.com
+    ///
+    /// - Parameter viewController: The view controller to be used as base for presenting the login flow.
+    /// - Returns: Nothing is returned for now. The client may return the profile infromation in the future.
+    func login(on viewController: UIViewController) async throws
+
+    /// Returns `true` if the client is authenticated.
+    func restoreState() async throws -> Bool
```
mohamede1945, line 54: I'd use `AuthenticationState` rather than bool.

Data/AuthenticationClient/Sources/AuthenticationClient.swift
```diff
@@ -0,0 +1,65 @@
...
+    func authenticate(request: URLRequest) async throws -> URLRequest
+
+    var authenticationState: AuthenticationState { get async }
+}
+
+public enum AuthentincationClientBuilder {
+    public static func make(withConfigurations config: AuthenticationClientConfiguration) -> AuthenticationClient {
+        AuthenticationClientImpl(configurations: config)
+    }
+}
```
mohamede1945, line 65: I'm not sure why would we need to use this over just AuthenticationClientImpl's initializer.

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
```diff
@@ -0,0 +1,302 @@
...
+
+    // MARK: Private
+
+    private var sut: AuthenticationClientImpl!
+    private var oauthService: OAuthServiceMock!
+    private var encoder: OAuthStateDataEncoder!
+    private var persistence: PersistenceMock!
+}
+
+private struct OauthStateEncoderMock: OAuthStateDataEncoder {
```
mohamede1945, line 185: I don't think we need this. I think we can use the real implementation.

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
```diff
@@ -0,0 +1,302 @@
...
+
+    func refreshIfNeeded(data: any OAuthStateData) async throws -> any OAuthStateData {
+        guard let refreshResult else {
+            fatalError()
+        }
+        return try refreshResult.get() ?? data
+    }
+}
+
+private final class AutehenticationDataMock: Equatable, Codable, OAuthStateData {
```
mohamede1945, line 249: Same here, can we get rid of this please?

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
```diff
@@ -0,0 +1,302 @@
...
+    var isAuthorized: Bool {
+        accessToken != nil
+    }
+
+    static func == (lhs: AutehenticationDataMock, rhs: AutehenticationDataMock) -> Bool {
+        lhs.accessToken == rhs.accessToken
+    }
+}
+
+private final class PersistenceMock: Persistence {
```
mohamede1945, line 281: Same for this mock as well.

mohamede1945 (changes requested):
Jazak Allah khyran for the changes. This looks so great mashaa'Allah, just a couple more more changes and we should be good to merge inshaa'Allah.

Core/OAuthService/OAuthService.swift
mohannad-hassan: Sounds about right. That's basically the gist of the underlying data.

Data/AuthenticationClient/Sources/AppAuthOAuthService.swift
mohannad-hassan: Good point. I missed this when I extracted the package out.

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
mohannad-hassan: The encoder is tied to the `OAuthStateData` implementation. `AppAuthStateData` encapsulates an instance of `OIDAuthState` that is not easy to control for tests.

Data/AuthenticationClient/Sources/AuthenticationClient.swift
mohannad-hassan: Sort of my habit: builders go hand in hand with protocols. Also, I was anticipating some logic before I made the client take non-nil configurations.

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
mohamede1945: Oh I see. Can we move the encode and decode methods to OAuthStateData protocol then? Something like:

```swift
public protocol OAuthStateData {
    func init(_ data: Data) throws
    func encode(_ data: OAuthStateData) throws -> Data
    ...
}
```

WDYT?

mohannad-hassan: We'll need to avoid hardcoding the concrete type, since `AuthenticationClient` is responsible for persisting the data.  If we go by the suggestion, it'll be something like this: 

```swift
public protocol OAuthStateData {
    func init(_ data: Data) throws
    func encode(_ data: OAuthStateData) throws -> Data
}

public protocol OAuthService {
    /// Concrete type can be accessed here but still abstracted. 
    static var stateDataType: OAuthStateData.Type { get }
    // ... 
}

func persistencePoint() {
     // How it'd be invoked. 
     let stateData = try currentAuthClient.Self.stateDataType.init(encodedData)
}
```

Honestly, I would only go for meta types in a generic function/class, so that's why I didn't think of this. 

This is more concise than using a separate encoder, but I think the encoder is simpler. So what do you think? @mohamede1945

Core/OAuthService/OAuthService.swift
```diff
@@ -0,0 +1,64 @@
...
+//  OAuthService.swift
+//  QuranEngine
+//
+//  Created by Mohannad Hassan on 08/01/2025.
+//
+
+import Foundation
+import UIKit
+
+public struct OAuthServiceConfiguration {
```
mohamede1945, line 11: I don't think this needs to be part of the API, does it?

Core/SystemDependencies/KeychainAccess.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  KeychainAccess.swift
+//  QuranEngine
+//
+//  Created by Mohannad Hassan on 21/01/2025.
+//
+
+import Foundation
+
+public protocol KeychainAccess {
```
mohamede1945, line 10: This is perfect! Thank you so much

Data/AuthenticationClient/Sources/AuthentincationClientImpl.swift
```diff
@@ -0,0 +1,166 @@
...
+            oauthService: service,
+            encoder: encoder,
+            persistence: KeychainPersistence()
+        )
+    }
+}
+
+private extension AuthenticationClientConfiguration {
+    // The interfaces for the configurations of both modules will change.
+    // Noticeably,
```
mohamede1945, line 157: Could you please explain why is that?

Example/QuranEngineApp/Classes/Container.swift
```diff
@@ -36,11 +36,11 @@ class Container: AppDependencies {
...
-        let client = AppAuthOAuthClient()
-        if let config = Constant.QuranOAuthAppConfigurations {
-            client.set(appConfiguration: config)
+    private(set) lazy var authenticationClient: (any AuthenticationClient)? = {
+        guard let configurations = Constant.QuranOAuthAppConfigurations else {
+            return nil
         }
+        let client = AuthenticationClientImpl(configurations: configurations)
         return client
     }()
```
mohamede1945, line 45: nit: This could be simply equals nil and remove the Constant.QuranOAuthAppConfigurations.

Package.swift
```diff
@@ -129,6 +129,19 @@ private func coreTargets() -> [[Target]] {
...
             .product(name: "AsyncAlgorithms", package: "swift-async-algorithms"),
         ]),
+
+        target(type, name: "OAuthService", hasTests: false, dependencies: []),
+
+        target(type, name: "SecurePersistence", hasTests: false, dependencies: [
+            "SystemDependencies",
+        ]),
+
+        target(type, name: "AppAuthOAuthService", hasTests: false, dependencies: [
```
mohamede1945, line 139: I would prefer naming this target OAuthServiceImpl or something similar, similar to how you named the fake to make it easy to find the API and both implementations.

mohamede1945 (approved):
Thanks Mohannad a lot for such wonderful work.

Core/OAuthService/OAuthService.swift
mohannad-hassan: It's practically a Parameters Object.

Data/AuthenticationClient/Sources/AuthentincationClientImpl.swift
mohannad-hassan: Forgot to complete the comment. Will commit it. Here it is: 

Noticeably, AuthenticationClient may accept an enum defining the available services. The client may request offline access and profile scopes by default. The OAuth service would still only accept String scopes.

On another hand, the issuer host is probably going to be the API host. We may see how the relationship pans out.

Example/QuranEngineApp/Classes/Container.swift
mohannad-hassan: I meant to leave the constants to be filled if any developer of a forked repo wanted to add their own OAuth account.

Data/AuthenticationClient/Sources/AuthentincationClientImpl.swift
mohannad-hassan: This is also meant to support forked apps. The OAuth account used by the apps is going to be automatically granted the requested scopes. 

I don't plan to put much effort in supporting this. I may just expose a property that returns the granted scopes.

Package.swift
mohannad-hassan: Okay, but I'll keep the `AppAuth`, as `OAuthServiceAppAuthImpl`.

Data/AuthenticationClient/Tests/AuthenticationClientTests.swift
mohamede1945: Sounds good, yeah let's keep it.

Core/SystemDependenciesFake/SecurityAccessFake.swift
```diff
@@ -0,0 +1,53 @@
...
+//  SecurityAccessFake.swift
+//  QuranEngine
+//
+//  Created by Mohannad Hassan on 21/01/2025.
+//
+
+import Foundation
+import SystemDependencies
+
+public class SecurityAccessFake: KeychainAccess {
```
mohamede1945, line 11: Could you please rename it to KeychainAccessFake?

mohamede1945 (approved):

PR #682, opened 2025-01-28
Page Bookmarks – Model Migrations

mohannad-hassan:
This PR introduces the needed changes in the page bookmarks table: adding an optional `remoteID`. This will be used to identify the bookmarks in APIs with the backend. 

Additionally, the PR enables automatic data migrations for the model. The changes are simple, so the automatic migration is sufficient. 

*Not in this PR*: the tracking of local mutations; either the table that tracks unsynced mutations or the any model or service changes.


mohamede1945:
I'm not sure what is this PR for. Could you please add description? 

Also theoretically, migration should be at the final stages when we develop a robust sync layer with quran.com, but maybe the description could clarify what this is about.

Thank you!

mohannad-hassan:
> I'm not sure what is this PR for. Could you please add description?
> 
> Also theoretically, migration should be at the final stages when we develop a robust sync layer with quran.com, but maybe the description could clarify what this is about.
> 
> Thank you!

@mohamede1945 Updated. Let me know if anything is missing. 

Note that this migration step is needed right now. We might have to do further migration steps related to the synchronisation logic.

PR #683, opened 2025-02-01
PageBookamrks – New Synced DB

mohannad-hassan:
First draft: shows a working example for a GRDB stack for page bookmarks.  Tests will be similar to the CoreData variant, except that the GRDB's integration with Combine doesn't guarantee that publishers will signal every time the resources is mutated, instead it might aggregate a few mutations together. 

Also, still remains to test the publisher function in `DatabaseConnection`.


Data/PageBookmarkPersistence/Sources/GRDBPageBookmarkPersistence.swift
```diff
@@ -0,0 +1,112 @@
...
+        }
+    }
+
+    private var migrator: DatabaseMigrator {
+        var migrator = DatabaseMigrator()
+        migrator.registerMigration("createPageBookmarks") { db in
+            try db.create(table: GRDBPageBookmark.databaseTableName, options: .ifNotExists) { table in
+                // Don't think we need a separate local id column.
+                table.column("page", .integer).primaryKey()
+                table.column("remote_id", .text)
```
mohamede1945, line 75: I think This should be non-null, since PageBookmarks table would be the persistence for synchronized information and doesn't hold local mutations.

Data/PageBookmarkPersistence/Sources/GRDBPageBookmarkPersistence.swift
```diff
@@ -0,0 +1,112 @@
...
+//  Created by Mohannad Hassan on 31/01/2025.
+//
+
+import Foundation
+import Combine
+import GRDB
+import VLogging
+import SQLitePersistence
+
+public struct GRDBPageBookmarkPersistence: PageBookmarkPersistence {
```
mohamede1945, line 14: I don't think it needs to conform to `PageBookmarkPersistence`. `PageBookmarkPersistence` is basically the repository (the pattern) API. Now GRDBPageBookmarkPersistence is not going to be used directly, instead it will be just a data source. This allows us to be able to supply a tailored entity that includes the remote ID as part of it.

mohannad-hassan: I was actually operating on the design that it's going to replace bookmarks' CoreDataPersistence in the app. 

- At the launch of the next update, data is migrated from CoreData to this one. 
- `GRDBPageBookmarkPersistence` is going to be the source of truth as far the app's presentation is concerned. 
- Some other storage will be created to handle sync-related concerns. i.e. local changes, last synced date, ... This is most likely going to be transient; once synced, local changes are going to be wiped and last synced date updated.

mohamede1945: I see two things we need that would make `GRDBPageBookmarkPersistence` not applicable to be the public API:

1. `GRDBPageBookmarkPersistence` only holds synchronized data. We need an additional persistence layer to store offline mutations that haven't been synchronized with the server yet (e.g., `GRDBPageBookmarkMutationsPersistence`).  
2. We need a higher-level class to manage interactions between different data sources (`GRDBPageBookmarkPersistence`, `GRDBPageBookmarkMutationsPersistence`, and `Quran.com RemoteDataSource`). This is typically implemented using the repository pattern. See an example here:https://developer.android.com/topic/architecture/data-layer.

mohannad-hassan: I get your point, and it makes sense.

Data/SQLitePersistence/Sources/DatabaseConnection.swift
```diff
@@ -109,6 +110,20 @@ public final class DatabaseConnection: Sendable {
...
 
+    /// Creates a publisher that tracks changes in the results of database requests,
+    /// and notifies fresh values whenever the database changes
+    ///
+    /// The first value is notified when the publisher is created. Subsequent changes *may* get coalesced in notifications.
+    public func readPublisher<T>(_ block: @Sendable @escaping (Database) throws -> T) throws -> AnyPublisher<T, Error> {
+        // - ValueObservation may coalesce subsequent changes into a single notification
+        // - ValueObservation fetches a fresh value immediately after a change *is committed in the database*.
+        // - By default, delivers on the main thread.
+        ValueObservation
```
mohannad-hassan, line 121: @mohamede1945  See here and see the test below.

I've pulled the inner comments from the documentation of `ValueObservation`.

mohamede1945: I see. Thank you!

Data/SQLitePersistence/Tests/DatabaseConnectionTests.swift
```diff
@@ -41,6 +42,52 @@ class DatabaseConnectionTests: XCTestCase {
...
+            .eraseToAnyPublisher()
+
+        var assertExpectation: XCTestExpectation?
+        var expectedNames: [String]?
+        let cancellable = publisher.sink { names in
+            guard let expected = expectedNames else { return }
+
+            if Set(expected) == Set(names) {
+                assertExpectation?.fulfill()
+                assertExpectation = nil
```
mohamede1945, line 62: I'm not quite sure how this test case passes. The closure captures the value of `assertExpectation`, so modifying it inside the closure wouldn't reflect outside of it, and vice versa—modifying the value outside the closure wouldn't be reflected inside. Am I missing something here? :)

mohamede1945: The thing here closure captures a reference to assertExpectation, but then later the reference itself is modifying to a new reference, so how come the closure is able to capture the new reference?

Data/SyncedPageBookmarkPersistence/Sources/SyncedPageBookmarkPersistence.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Created by Mohannad Hassan on 09/02/2025.
+//
+
+import Combine
+import Foundation
+
+public protocol SyncedPageBookmarkPersistence {
+    func syncedPageBookmarksPublisher() throws -> AnyPublisher<[SyncedPageBookmarkPersistenceModel], Never>
+    func insert(bookmark: SyncedPageBookmarkPersistenceModel) async throws
+    func removeBookmark(withRemoteID remoteID: String) async throws
```
mohamede1945, line 14: I think we might need to make these consistent, I suggest we rename them to:

```
func pageBookmarksPublisher() throws -> AnyPublisher<[SyncedPageBookmarkPersistenceModel], Never>
func insertBookmark(_ bookmark: SyncedPageBookmarkPersistenceModel) async throws
func removeBookmark(withRemoteID remoteID: String) async throws
```

mohamede1945, line 14: For Sync'd bookmarks, I believe `page` should be the key not the remoteID. Hence, I suggest the removeBookmark to take the page instead of remote Id.

Package.swift
```diff
@@ -196,20 +196,32 @@ private func uiTargets() -> [[Target]] {
...
-        // MARK: - Core Data
+        // MARK: - Page Bookmarks
 
-        target(type, name: "LastPagePersistence", dependencies: [
+        target(type, name: "PageBookmarkPersistence", dependencies: [
             "CoreDataModel",
             "CoreDataPersistence",
+            "SQLitePersistence",
+            "QuranKit",
+            .product(name: "GRDB", package: "GRDB.swift"),
```
mohamede1945, line 206: Please revert

Package.swift
```diff
@@ -196,20 +196,32 @@ private func uiTargets() -> [[Target]] {
...
         ], testDependencies: [
             "AsyncUtilitiesForTesting",
             "CoreDataPersistenceTestSupport",
         ]),
 
-        target(type, name: "PageBookmarkPersistence", dependencies: [
+        target(type, name: "SyncedPageBookmarkPersistence", dependencies: [
+            "SQLitePersistence",
+            "QuranKit",
```
mohamede1945, line 214: I don't think this target depends on QuranKit

mohamede1945 (changes requested):
Amazing work, Mohannad! I like the test cases a lot too. Just a couple of small changes and we should be good to go inshaa'Allah.

Data/SQLitePersistence/Tests/DatabaseConnectionTests.swift
mohannad-hassan: I often like to utilize around with nested functions whenever there's a room. So your comment actually gave me little fright, because I wasn't sure of the behavior of escaping closures. If a capturing closure is returned or returned as an escaping closure, the compiler moves the variables it captures implicitly to the heap. 

It seems to me that warnings against capturing self implicitly in closures is a special thing for self.

Data/SyncedPageBookmarkPersistence/Sources/SyncedPageBookmarkPersistence.swift
mohannad-hassan: Oh I see. 

The "synced" part is implied from the type's name anyways.

mohannad-hassan: How about we add a `unique` constraint on `page` and keep the primary key for `remote_id`? 

This way we can keep the two ideas; signaling that `page` is unique and that the design of the synced database follows the server's design. 

BTW, this will be different in both the local database and the app's model.

mohamede1945: Ok, yes a unique constraint could help here. Thank you!

mohamede1945 (approved):

PR #688, opened 2025-03-07
Fix some Kazakh localization

abdirakhman-wwbw:


mohamede1945:
Jazak Allah Khyrn for the review and updates.

PR #696, opened 2025-03-22
Put Login with Quran.com behind a compiler flag

mohamede1945:
We can add `-DQURAN_SYNC` to "other swift flags" to enable it.


mohamede1945:
@mohannad-hassan FYI

mohannad-hassan (approved):

PR #700, opened 2025-04-02
Add Surah Sorting to Home View

yismailuofa:
### Summary
I thought it would be pretty useful to sort the surahs if you want to read from the other side, this is my first crack at it.

### Demo
https://github.com/user-attachments/assets/aa430e66-f35a-46b1-beeb-158a002d3c1c

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,9 @@ jobs:
       run: sudo xcode-select -s "/Applications/Xcode_16.2.app"
 
     - name: Run tests
-      run: set -o pipefail && xcrun xcodebuild build test -scheme QuranEngine-Package -sdk "iphonesimulator" -destination "name=iPhone 16,OS=18.2" | xcpretty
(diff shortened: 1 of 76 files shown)


yismailuofa:
unsure if the CI error is related to my change 🤔

mohamede1945:
Jazak Allah khyrn for the PR. It looks great. Just one comment, I'm unsure if we should save this sorting. IMO, I would not save it since I think this looks like a way to easily navigate to the bottom of the list rather than a new navigation preference for users.

PS: https://github.com/quran/quran-ios/pull/702 should resolve the build issue.

yismailuofa:
agreed; took out the user defaults and refactored the code abit to move the sorting into one place

mohamede1945 (approved):

mohamede1945:
Jazak Allah khyrn, Youssef

yismailuofa:
salam, is it possible to get this merged into the app store version?

PR #708, opened 2025-07-25
Misc: AppIcon for example app & setting other settings for packages

mohannad-hassan:
## Summary by CodeRabbit

* **New Features**
  * Added support for specifying additional Swift compiler settings when defining a target.


mohannad-hassan:
Oh sorry about that.

For the app icon, I needed to remove the alpha channel so that it can be uploaded to the App Store. I honestly used an online service to quickly change it, and white seemed non-noisy for me. Since we might need to upload builds for tests, I thought we might keep it. 

For the other settings, I added that to pass the other compiler flags, which will be used to enable the flag for Quran sync work.

mohamede1945:
No worries, Mohannad! 

I'd typically lean toward including the settings change in the same PR where it’s used, to keep the change atomic. But it’s fine this time :)

PR #713, opened 2025-11-01
Audio speed control (new feature)

adnan29979:
I added the speed control on the iOS Audio Banner, similar to what exists on the Web. I think Android has it too (not 100% sure). I don’t have an Android phone right now to test, but I saw older PRs/MRs that suggest it’s there. It felt missing on iPhone, so I built it.

I moved the Speed button to the left side, right after Stop. Then it goes Backward → Play/Pause → Forward → More. This is all done in SwiftUI, so it’s not hard-coded with frames. It’s just the order in the HStack in AudioBannerViewUI.swift (the AudioPlaying view). The layout is dynamic and follows the view code.

I set 1× as the default speed. If the user has no saved setting, we seed UserDefaults("Audio.PlaybackRate") with 1.0 on first run and apply it to the player on start. If they change speed from the banner, I write the new value back to UserDefaults, so it sticks for next time.
image
image

### DIFF
--- a/Core/QueuePlayer/AudioPlayer.swift
+++ b/Core/QueuePlayer/AudioPlayer.swift
@@ -206,7 +206,7 @@ class AudioPlayer {
     }
 
     private func waitUntilFrameEnds(currentTime: TimeInterval? = nil) {
-        // Remaining media time to the end of the frame 
+        // max with 100ms since sometimes the returned value could be negative
         let mediaDelta = max(0, getDurationToFrameEnd(currentTime: currentTime))
         // Convert media time to wall-clock time 
         let rate = max(0.1, Double(player.effectiveRate))


adnan29979:
@mohamede1945 , can you kindly please share a rough timeline when you'd review this change and provide a feedback or merge it (if all seems okay)?
I know you might have other jobs which can delay this review. but approximately when can I expect a review please?
Thanks.

UI/NoorUI/Features/AudioBanner/AudioBannerViewUI.swift
```diff
@@ -25,7 +25,24 @@ public struct AudioBannerActions {
     let cancelDownloading: AsyncAction
     let reciters: () -> Void
     let more: () -> Void
-    public init(play: @escaping () -> Void, pause: @escaping () -> Void, resume: @escaping () -> Void, stop: @escaping () -> Void, backward: @escaping () -> Void, forward: @escaping () -> Void, cancelDownloading: @escaping AsyncAction, reciters: @escaping () -> Void, more: @escaping () -> Void) {
+    // NEW
```
mohamede1945: Do you mind deleting these kind  of comments, please?

Core/QueuePlayer/QueuePlayer.swift
```diff
@@ -75,6 +81,7 @@ public class QueuePlayer {
             oldValue?.actions = nil
         }
     }
+    private var playbackRate: Float = 1.0
```
mohamede1945: Since the rate is persisted and also audio player stores it, I don't think we need this property.

UI/NoorUI/Features/AudioBanner/AudioBannerViewUI.swift
```diff
@@ -65,6 +85,23 @@ public struct AudioBannerViewUI: View {
...
+    ("1.25×", 1.25), ("1.5×", 1.5), ("1.75×", 1.75), ("2×", 2.0),
+]
+
+private func speedLabel(_ rate: Float) -> String {
+    if let m = kSpeedOptions.first(where: { abs($0.value - rate) < 0.001 }) {
+        return m.title
+    }
+    var s = String(format: "%.2f", rate)
+    while s.contains(".") && (s.hasSuffix("0") || s.hasSuffix(".")) { s.removeLast() }
+    return s + "×"
```
mohamede1945: Why do we need to do this?

mohamede1945 (changes requested):
Jazak Allah khyrn for the PR.  I've left a couple of comments.

I tested it locally and the major issue I see is that when I change the speed, the recitation highlighting doesn't work.

Core/QueuePlayer/Player.swift
```diff
@@ -46,7 +48,7 @@ final class Player {
     }
 
     func play() {
-        player.play()
+        player.playImmediately(atRate: currentRate)
```
adnan29979: line 22, 24, 51 is because at higher speeds, when the audio transitions from one ayah to the other, there is a micro gap of silence. It brings down the user experience. These lines take care of that 'dips' between ayahs at speeds other than 1x

adnan29979:
AssalamuAlaikum. Is there any way I can help bring any changes to this PR to finalize it? Thanks.

mohamede1945 (approved):

Issue #691, opened 2025-03-17
Allow infinite scrolling instead of pages


basememara:
If I want to get to ayah 150 of a surah, seems like I have to flip page by page instead of having an infinite scroll where I can scrub the scroller.

mohamede1945:
This is in the translation view not the Arabic view, right?

adnan29979:
@mohamede1945 in translation view, is this a feature worth adding?
If so, I'd be happy to start working on it.

Edit: and which way would be the best in implementing it?

PR #736, opened 2025-12-10
New Feature: Vertical Scroll

adnan29979:
QuranVerticalPaginationView.swift for vertical scrolling in translation mode. 
MoreMenuScrollDirection.swift for toggle between horizontal and vertical scrolling

notes:
loadedRange always starts from page 0 so that it can only only expand downward. This avoids a jarring scroll jump bug. 

When ContentTranslationViewBody is loading (the items are empty), it shows a ProgressView with a very light/transparent background. At the same time, QuranVerticalPaginationView always draws PageMiddleSeparator() between pages. Because of that, you could see white vertical lines appear while the pages were loading. Just a small user experience bug. 
So we gave the loading state a proper background color (matching the app theme) so those separators stay hidden until the content is ready.

Please feel free to ask for changes if required. I have tested this vigorously, and there are no issues from my testing so far.


adnan29979:
#691

mohamede1945:
I think the biggest problem is the 2 cascaded scroll views. The experience will not be optimal when the users can reach the end of the page, they have to lift their finger and then scroll again to go to the next page rather than having just a single continuous scroll for all the ayat

adnan29979:
I may be misunderstanding your concern, because continuous scrolling is already implemented. In vertical mode, we set scrollingEnabled = false on each page, which makes ContentTranslationView use a plain LazyVStack instead of a List.

Since LazyVStack has no scrolling of its own, there's only one ScrollView (in QuranVerticalPaginationView) that handles all scrolling. Users should be able to scroll continuously through all ayat without lifting their finger.

Could you kindly clarify what you are experiencing?

adnan29979:
Assalamualaikum. Thanks. Any feedback you'd like to kindly share after testing?

PR #751, opened 2026-02-26
Add QUL Naskh mushaf font support

rayinuk13:


mohamede1945:
@rayinuk13 thanks for putting up the PRs.

However, this change is not sufficient. Adding the font file and a helper function alone does not integrate it into the app. The font is currently not being used anywhere in the UI.

Additionally, we still need the Naksh Mushaf images to be included and properly wired into the rendering layer. Without those assets, the feature is incomplete.

Did you run the example app locally and verify the visual changes?

PR #784, opened 2026-04-04
Avoid removing pages when possible

ahmedre:
Sometimes, when the version of Readings is updated, the mobile phone
already has the newest version (because of the time delta between when
the readings were updated on the server and when they are updated in the
app). This patch lets the app first check for the marker file - if it's
there, there is no need to remove and re-download, since the current
version of the files is already present.

### DIFF
--- a/Domain/ReadingService/Sources/ReadingRemoteResources.swift
+++ b/Domain/ReadingService/Sources/ReadingRemoteResources.swift
@@ -19,6 +19,7 @@ public struct RemoteResource {
 
     public init(url: URL, reading: Reading, version: Int) {
         self.url = url
+        self.reading = reading
         self.version = version
         downloadDestination = Reading.readingsPath.appendingPathComponent(reading.localPath, isDirectory: true)
     }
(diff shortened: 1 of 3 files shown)


ahmedre:
one caveat here pointed out by Codex - if a person successfully got the zip file and unzipping fails mid way through (not enough space, etc), the version file could exist without the files existing. this is especially true since we don't try to remove partial unzipped content if unzip fails, etc.

so this would kind of make a certain broken behavior worse (person failed unzip, but did get the `.v2` file, for example) - before this pr, we'd do nothing, and, on next open, there's no `success-v2.txt`, so we remove the folder and try again. but today, this patch could possibly find `.v2` file and say that we're good. I still think this is worth having.

maybe one thought is if we have the .v2 file but don't have `success-v2.txt`, we do a check to ensure every page exists and if so, write the `success-v2.txt` file, otherwise don't - wdyt? alternatively (though this would only help new cases), remove the unzipped content if unzip fails.

ahmedre:
I did a check to work around this in #785 - basically on upgrade, iterate and ensure we have all the images.

mohamede1945 (approved):

PR #785, opened 2026-04-04
Check that files exist on image upgrade

ahmedre:
When the image version is upgraded within the app, it's likely that the
person using the app may already have the newest images. We can find out
by checking for the version file within the zip file's extractions. This
patch makes the logic safer by ensuring that zip files that got
extracted incompletely (due to some file system error, out of space,
etc) are not also presumed to be downloaded.

### DIFF
--- a/Domain/ReadingService/Sources/ReadingRemoteResources.swift
+++ b/Domain/ReadingService/Sources/ReadingRemoteResources.swift
@@ -28,6 +28,13 @@ public struct RemoteResource {
 
     public let downloadDestination: RelativeFilePath
 
+    public func isDownloaded(fileSystem: FileSystem = DefaultFileSystem()) -> Bool {
+        if hasSuccessFile(fileSystem: fileSystem) {
+            return true
+        }
+        return canRecoverSuccessFile(fileSystem: fileSystem)
+    }
+
     // MARK: Internal
 
     let url: URL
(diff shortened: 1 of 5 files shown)


ahmedre:
Requires #784.
This will be quite expensive if line by line images are ever updated though, since it checks 15 * 604 pages. on the bright side, once in sha' Allah we add the patch logic, we can potentially skip this (i.e. if we have _any_ `success_` file, just download the patch).

ahmedre:
I just thought of a good simplification - instead of this, we can do something else - 

roughly in kotlin-ish pseudocode:

```kotlin
val successFile = File(successPath, "success_$version.txt")
val versionFile = File(imagesPath, ".v$version")
return if (successFile.exists()) {
  true
} else if (versionFile.exists() {
   for v in 1 until version {
       val successVersionFile = File(successPath, "success_$v.txt")
       if (successVersionFile.exists()) {
           return true
       }
   }
}
return false
```

so if we have a `.vX` file, we need to also see _any_ `success_<>.txt` to know that we successfully unzipped a file. that would save us from 604 file checks and/or 604 * 15 file checks - wdyt?

incidentally, a future patch should say something like this for else -

```kotlin
else if (anyVersionFileExists()) {
   // download the patch instead of the full delete and full data set
}
```

mohamede1945:
Yeah, your last suggestion is great, let's do it. 

Also I'm surprised that I didn't remove the partially downloaded stuff [here](https://github.com/quran/quran-ios/blob/3ba7c20d627c4c2f1019636373aba1c55884c0db/Domain/ReadingService/Sources/ReadingResourcesService.swift#L153-L156). Would you mind doing this too?


BTW, thanks for the width refactor, looks much better

mohamede1945 (approved):

PR #786, opened 2026-04-04
Update settings auth flow and remove QuranProfileService

AhmedNMahran:
## Summary

This PR updates the Quran.com account area in Settings and removes the extra `QuranProfileService` layer.

The main goals are:
- move the Quran.com account section to the top of Settings
- update the signed-out and signed-in UI to match the new design direction while reusing existing NoorUI components
- localize the new Settings strings in English and Arabic
- open the Quran.com profile page from the signed-in settings row
- remove `QuranProfileService` and let Settings / Bookmarks use `AuthenticationClient` directly **was requested by @mohamede1945 in the previous PR as a follow-up**

### Settings UI
- moves the Quran.com account section to the top of the Settings list
- keeps the layout on existing NoorUI list building blocks:
  - `NoorList`
  - `NoorBasicSection`
  - `NoorListItem`
  - `NoorSystemImage`
- signed-out state now shows:
  - sign in row
  - sync across all devices
  - create custom collections
  - attach notes to verses
- signed-in state now shows:
  - the current user email when available
  - a profile/settings action that opens `https://quran.com/profile`
  - a sign-out row

### Localization
- adds localized Settings strings in:
  - `Core/Localization/Resources/en.lproj/Localizable.strings`
  - `Core/Localization/Resources/ar.lproj/Localizable.strings`
- removes hardcoded user-facing strings from the new Quran.com Settings section

### Remove `QuranProfileService`
- deletes the `QuranProfileService` target and its tests
- updates `SettingsRootViewModel` to use `AuthenticationClient` directly for:
  - restore state
  - login
  - logout
  - current user email
- updates `BookmarksViewModel` to use `AuthenticationClient` directly for auth state and login
- keeps the email retrieval on the auth-service-backed path through `AuthenticationClientMobileSyncImpl`

### Package / test updates
- updates `Package.swift` to remove `QuranProfileService` and wire feature targets directly to `AuthenticationClient`
- adds replacement coverage in:
  - `Features/SettingsFeature/Tests/SettingsRootViewModelTests.swift`
  - `Features/BookmarksFeature/Tests/BookmarksViewModelTests.swift`

## Validation

- `make format-lint`
- `swift package dump-package`
- package build through the Example project:
  - `xcodebuild -project Example/QuranEngineApp.xcodeproj -scheme QuranEngine-Package -destination 'generic/platform=iOS' build`

## Notes

- The sync-related Settings UI remains behind `QURAN_SYNC`.
- For local sync testing, package targets still need Xcode to be launched with `QURAN_SYNC=1` so the package-side compilation flag is applied.


Data/AuthenticationClient/Sources/AuthenticationClient.swift
```diff
@@ -63,4 +63,11 @@ public protocol AuthenticationClient {
     func getAuthenticationHeaders() async throws -> [String: String]
 
     var authenticationState: AuthenticationState { get async }
+    var currentUserEmail: String? { get async }
+}
+
+public extension AuthenticationClient {
+    var currentUserEmail: String? {
+        get async { nil }
+    }
```
mohamede1945, line 77: I don't think we should provide a default implementation here

Features/BookmarksFeature/BookmarksViewModel.swift
```diff
@@ -49,7 +49,16 @@ final class BookmarksViewModel: ObservableObject {
     }
 
     func start() async {
-        isAuthenticated = await quranProfileService.refreshAuthenticationState() == .authenticated
+        if let authenticationClient {
+            do {
+                isAuthenticated = try await authenticationClient.restoreState() == .authenticated
+            } catch {
+                isAuthenticated = await authenticationClient.authenticationState == .authenticated
+            }
```
mohamede1945: I would add a convenience to `AuthenticationClient`

extension AuthenticationClient {
  func safelyRestoreState() -> AuthenticationState { // put the code above here }
}

Data/AuthenticationClient/Sources/AuthenticationClient.swift
```diff
@@ -63,4 +63,11 @@ public protocol AuthenticationClient {
     func getAuthenticationHeaders() async throws -> [String: String]
 
     var authenticationState: AuthenticationState { get async }
+    var currentUserEmail: String? { get async }
```
mohamede1945: Instead of user's email, I would expose directly the user's model. For example: `var loggedInUser: UserInfo?`

Data/AuthenticationClient/Sources/AuthenticationClientMobileSyncImpl.swift
```diff
@@ -16,6 +16,10 @@ public final actor AuthenticationClientMobileSyncImpl: AuthenticationClient {
         authService.isLoggedIn() ? .authenticated : .notAuthenticated
     }
 
+    public var currentUserEmail: String? {
+        (authService.authState as? AuthState.Success)?.userInfo.email
```
mohamede1945: Isn't authState an enum? I think we should add to the `mobile-sync-spm` repo, this so that we get a better API:

```swift
public enum AuthStateValue {
  case idle
  case loading
  case success(UserInfo)
  case error(exception: KotlinException?, message: String)

  public init(_ state: AuthState) {
    switch state {
    case is AuthStateIdle:
      self = .idle

    case is AuthStateLoading:
      self = .loading

    case let success as AuthStateSuccess:
      self = .success(success.userInfo)

    case let error as AuthStateError:
      self = .error(
        exception: error.exception as? KotlinException,
        message: error.message
      )

    default:
      fatalError("Unhandled AuthState subtype: \(type(of: state))")
    }
  }
}

extension AuthService {
  var authStateValue: AuthStateValue { AuthStateValue(authState) }
}
```

Which would allow us to implement something like:

```swift
var user: UserInfo? {
  switch authService.authStateValue: {
  case .success(let user): return user
  case .idle, .loading, .error: return nil
}
```

Core/Localization/Resources/en.lproj/Localizable.strings
```diff
@@ -52,6 +52,12 @@
 "setting.donate" = "Donate";
 "setting.share_app" = "Share the app...";
 "setting.app_version" = "App Version";
+"setting.quran_account.sign_in" = "Sign in";
+"setting.quran_account.profile" = "Quran.com Profile";
+"setting.quran_account.sign_out" = "Sign out";
+"setting.quran_account.sync_devices" = "Sync across all devices";
+"setting.quran_account.custom_collections" = "Create custom collections";
+"setting.quran_account.attach_notes" = "Attach notes to verses";
```
mohamede1945: I think we agreed not to have these behind login. They will be available to non-logged in users too

Features/SettingsFeature/SettingsRootViewModel.swift
```diff
@@ -136,35 +136,56 @@ final class SettingsRootViewModel: ObservableObject {
...
+            isAuthenticated = false
+            currentUserEmail = nil
+            return
+        }
+
+        do {
+            isAuthenticated = try await authenticationClient.restoreState() == .authenticated
+        } catch {
+            isAuthenticated = await authenticationClient.authenticationState == .authenticated
+        }
```
mohamede1945: Let's use the convenience method I suggest above please

Package.swift
```diff
@@ -619,15 +615,35 @@ private func featuresTargets() -> [[Target]] {
...
+                    .init(stringLiteral: "BookmarksFeature"),
+                    .init(stringLiteral: "Analytics"),
+                    .init(stringLiteral: "AnnotationsService"),
+                    .init(stringLiteral: "AuthenticationClient"),
+                    .init(stringLiteral: "PageBookmarkPersistence"),
+                ],
+                path: "Features/BookmarksFeature/Tests",
+                swiftSettings: settings
+            ),
+        ],
```
mohamede1945: This is not how we add tests. Revert to the original imp, then remove `hasTests: false` and use `testDependencies`. Same for settings feature

mohamede1945 (changes requested):
Jazak Allah khyrn for the changes!

AhmedNMahran:
@mohamede1945 Jazak Allah Khayran for the review
kindly check after applying the changes

Updated the PR with the requested cleanup and API follow-up.

Changes in this update:
- bumped `mobile-sync-spm` to `0.0.6`
- switched `quran-ios` to use the new `AuthService.loggedInUser` API instead of unpacking raw auth state in app code
- removed `QuranProfileService` and moved the remaining auth handling directly to the feature view models
- kept the `AuthenticationClient` app boundary using a local `LoggedInUser` model rather than exposing the MobileSync/KMP model directly
- added `AuthenticationClient.safelyRestoreState()` and reused it from Settings/Bookmarks
- updated the Settings card/profile flow
- made the profile URL environment-aware:
  - prod: `https://quran.com/profile`
  - prelive: `https://prelive.quran.com/profile`

Validation:
- `make format-lint`
- package tests via `xcodebuild build test -scheme QuranEngine-Package`
- example app build via `xcodebuild build -project Example/QuranEngineApp.xcodeproj -scheme QuranEngineApp`
- sync-enabled app build + simulator launch locally

Data/AuthenticationClient/Sources/AuthenticationClient.swift
```diff
@@ -24,6 +25,42 @@ public enum AuthenticationState: Equatable {
     case authenticated
 }
 
+public struct LoggedInUser: Equatable, Sendable {
```
mohamede1945: I don't think we need this struct. We can use UserInfo directly.

I've approved the PR. Feel free to do the change and merge directly.

Thanks ya Ahmed!

Core/Localization/Resources/en.lproj/Localizable.strings
mohamede1945: reminder

mohamede1945 (approved):

Core/Localization/Resources/en.lproj/Localizable.strings
AhmedNMahran: @mohamede1945 could you please elaborate? what I understood from your previous comment is you need to change wording so that is doesn't imply that the bookmarks/notes/highlights are enabled on login only.

if you mean something else please let me know.

Data/AuthenticationClient/Sources/AuthenticationClient.swift
AhmedNMahran: Thanks @mohamede1945, I added the model so that we don't force the direct usage of the mobile-sync model, from non-sync feature. but I can remove it no issues, Jazak Allah Khayran

Core/Localization/Resources/en.lproj/Localizable.strings
mohamede1945: I think we simply want to remove them. I believe Talha has removed them from the design while we were in the meeting. You could double check with him.

AhmedNMahran: They're still there, those are for the settings screen, the bookmarks page should have the nudge.
so I think we're ok to leave them.

mohamede1945: I saw you changed the text to read about sync'ing not functionality in general. So we are good, my bad

PR #788, opened 2026-04-09
Remove partially unzipped files

ahmedre:


ahmedre:
note  there's an intentional decision I made here, which is to also erase the files if the download and success succeed but the success-vX.txt file fails to write.

mohamede1945 (approved):

PR #789, opened 2026-04-09
Add bookmark collection sync wrapper

AhmedNMahran:
## Summary
Add a thin Mobile Sync collection wrapper only.

## Scope
- add `MobileSyncSupport`
- add `BookmarkCollectionService`
- wire it through `AppDependencies` / `Container` under `QURAN_SYNC`

## Out of scope
- no highlights UI
- no ayah menu changes
- no Quran page rendering changes
- no note/highlight behavior changes

## Notes
This keeps the new sync integration isolated from older modules and avoids mixing data and UI in the same PR.

### DIFF
--- a/Example/QuranEngineApp/Classes/Container.swift
+++ b/Example/QuranEngineApp/Classes/Container.swift
@@ -47,6 +47,10 @@ class Container: AppDependencies {
     }()
 
     private(set) lazy var notePersistence: NotePersistence = CoreDataNotePersistence(stack: coreDataStack)
+    #if QURAN_SYNC
+        private(set) lazy var syncService: SyncService? = mobileSyncServices?.syncService
+    #endif
+
     private(set) lazy var authenticationClient: (any AuthenticationClient)? = {
         #if QURAN_SYNC
             if let authService = mobileSyncServices?.authService {
--- a/Features/AppDependencies/AppDependencies.swift
+++ b/Features/AppDependencies/AppDependencies.swift
@@ -11,6 +11,10 @@ import AuthenticationClient
 import BatchDownloader
 import Foundation
 import LastPagePersistence
+#if QURAN_SYNC
+    import MobileSync
+    import MobileSyncSupport
+#endif
 import NotePersistence
 import PageBookmarkPersistence
 import QuranResources
@@ -38,6 +42,11 @@ public protocol AppDependencies {
     var notePersistence: NotePersistence { get }
     var pageBookmarkPersistence: PageBookmarkPersistence { get }
 
+    #if QURAN_SYNC
+        var syncService: SyncService? { get }
+        var bookmarkCollectionService: BookmarkCollectionService? { get }
+    #endif
+
     var authenticationClient: (any AuthenticationClient)? { get }
 }
 
@@ -50,6 +59,13 @@ extension AppDependencies {
             quranFileURL: quranUthmaniV2Database
         )
     }
+
+    #if QURAN_SYNC
+        public var bookmarkCollectionService: BookmarkCollectionService? {
+            syncService.map { BookmarkCollectionService(syncService: $0) }
+        }
+
+    #endif
 
     public func noteService() -> NoteService {
         NoteService(
--- a/Data/MobileSyncSupport/Sources/BookmarkCollectionService.swift
+++ b/Data/MobileSyncSupport/Sources/BookmarkCollectionService.swift
@@ -1,0 +1,54 @@
+//
+//  BookmarkCollectionService.swift
+//
+//
+//  Created by Ahmed Nabil on 2026-04-22.
+//
+
+#if QURAN_SYNC
+    import Foundation
+    import MobileSync
+    import QuranKit
+
+    public struct BookmarkCollectionService {
+        // MARK: Lifecycle
+
+        public init(syncService: SyncService) {
+            self.syncService = syncService
+        }
+
+        // MARK: Public
+
+        public func createCollection(named name: String) async throws {
+            try await syncService.createCollection(named: name)
+        }
+
+        @discardableResult
+        public func addAyahBookmark(_ ayah: AyahNumber, toCollectionLocalId collectionLocalId: String) async throws -> Bookmark {
+            let bookmark = try await syncService.addAyahBookmark(
+                sura: Int32(ayah.sura.suraNumber),
+                ayah: Int32(ayah.ayah)
+            )
+            try await addBookmark(bookmark, toCollectionLocalId: collectionLocalId)
+            return bookmark
+        }
+
+        public func addBookmark(_ bookmark: Bookmark, toCollectionLocalId collectionLocalId: String) async throws {
+            try await syncService.addBookmarkToCollection(
+                collectionLocalId: collectionLocalId,
+                bookmark: bookmark
+            )
+        }
+
+        public func removeBookmark(_ bookmark: Bookmark, fromCollectionLocalId collectionLocalId: String) async throws {
+            try await syncService.removeBookmarkFromCollection(
+                collectionLocalId: collectionLocalId,
+                bookmark: bookmark
+            )
+        }
+
+        // MARK: Private
+
+        private let syncService: SyncService
+    }
+#endif
--- a/Package.swift
+++ b/Package.swift
@@ -229,6 +229,11 @@ private func dataTargets() -> [[Target]] {
             "AsyncUtilitiesForTesting",
         ]),
 
+        target(type, name: "MobileSyncSupport", hasTests: false, dependencies: [
+            "QuranKit",
+            .product(name: "MobileSync", package: "mobile-sync-spm"),
+        ]),
+
         // MARK: - Core Data
 
         target(type, name: "LastPagePersistence", dependencies: [
@@ -538,6 +543,8 @@ private func featuresTargets() -> [[Target]] {
             "ReadingService",
             "QuranResources",
             "AuthenticationClient",
+            "MobileSyncSupport",
+            .product(name: "MobileSync", package: "mobile-sync-spm"),
         ]),
 
         target(type, name: "FeaturesSupport", hasTests: false, dependencies: [


AhmedNMahran:
@mohamede1945 
I split this into 3 stacked PRs to keep review scope tighter:

1. PR 1: data/behavior foundation
   - separates highlights from notes
   - adds the sync-backed highlight path
   - highest-risk behavioral change

2. PR 2: Bookmarks feature UI
   - adds Highlights entry + list/detail screens
   - easier to review once the backend shape is settled

3. PR 3: presentation polish
   - shared UI components, spacing/alignment, note/highlight presentation
   - intended to be mostly visual / low-risk on top of the earlier behavior changes

This kept the risky data-model change separate from the feature UI and from the final polish pass.

Core/Localization/Resources/ar.lproj/Localizable.strings
mohamede1945, line 1: I don't think we should modify existing localization to repurpose it.

1. This doesn't update all translations we have
2. We should be running the new screens along side the existing screens and control which ones to show based on QURAN_SYNC flag

Data/HighlightPersistence/Sources/HighlightPersistence.swift
mohamede1945, line 1: 1. We need AyahCollectionPersistence rather than HighlightPersistence. 
2. HighlightPersistence can wrap AyahCollectionPersistence, though I wouldn't do that. I would not have HighlightPersistence at all, this can just be a feature detail that certain named collections will be colored rather than persistence given persistence and data model doesn't support the coloring.

Data/HighlightPersistence/Sources/MobileSyncHighlightPersistence.swift
```diff
@@ -0,0 +1,213 @@
...
+                }
+            } catch {
+                // Ignore sync errors in UI observation.
+            }
+        }
+
+        return subject
+            .handleEvents(receiveCancel: { task.cancel() })
+            .eraseToAnyPublisher()
+    }
```
mohamede1945: No need convert async sequence to publisher, that was the old code, now we can use async sequence directly when needed

Data/HighlightPersistence/Sources/MobileSyncHighlightPersistence.swift
```diff
@@ -0,0 +1,213 @@
...
+private extension QuranAnnotations.Note.Color {
+    static let syncedHighlightColors: [Self] = [.red, .green, .blue, .yellow, .purple]
+
+    init?(highlightCollectionName: String) {
+        switch highlightCollectionName.lowercased() {
+        case "red": self = .red
+        case "green": self = .green
+        case "blue": self = .blue
+        case "yellow": self = .yellow
+        case "purple": self = .purple
```
mohamede1945: I think the names here might not be accurate. Could you verify with the backend please? I think they have "ios" in the names.

Data/HighlightPersistence/Sources/PreferencesHighlightPersistence.swift
mohamede1945, line 1: What is this? I don't understand? Why would we save highlights as user defaults?

Domain/AnnotationsService/Sources/HighlightService.swift
```diff
@@ -0,0 +1,139 @@
...
+
+    // MARK: Public
+
+    public func color(from notes: [Note]) -> Note.Color {
+        notes.max { $0.modifiedDate < $1.modifiedDate }?.color ?? lastUsedHighlightColor
+    }
+
+    public func highlights(quran: Quran) -> AnyPublisher<[Note], Never> {
+        Task {
+            try? await migrateLegacyHighlightsIfNeeded()
```
mohamede1945: We've discussed this before, we shouldn't add migration now, this is complex feature and requirements can change and data model can change and we haven't finalized the requirements nor the impl yet so migration should be the last step

Domain/AnnotationsService/Sources/HighlightService.swift
```diff
@@ -0,0 +1,139 @@
+import Analytics
+import Combine
+import Foundation
+import HighlightPersistence
+import NotePersistence
+import Preferences
+import QuranAnnotations
+import QuranKit
+
+public struct HighlightService {
```
mohamede1945: I don't think we need this service for highlights, let's add it only if we are removing duplicate code in the UI PRs otherwise the logic can simply live in the view model

Domain/AnnotationsService/Sources/NoteService.swift
```diff
@@ -11,7 +11,6 @@ import Combine
 import Foundation
```
mohamede1945, line 11: Please keep the existing logic we shouldn't simply remove it, we need to maintain the existing behavior while keeping new behavior behind QURAN_SYNC

UI/NoorUI/Features/Note/HighlightPaletteIcon.swift
```diff
@@ -0,0 +1,41 @@
+import QuranAnnotations
+import SwiftUI
+import UIx
+
+public struct HighlightPaletteIcon: View {
```
mohamede1945: These are UI changes not data persistence. There are other UI changes too in this PR

UI/NoorUI/Features/Note/NoteEditorView.swift
```diff
@@ -23,18 +23,8 @@ public struct NoteEditorView: View {
 
     public var body: some View {
         VStack {
```
mohamede1945, line 25: As I mentioned before, we can't simply update existing functionality, we need to gate the changes behind QURAN_SYNC.

mohamede1945 (changes requested):
Thanks a lot for the PR. A couple of general comments:

1. I'm seeing UI changes in this PR, even though it's described as data layer changes only PR

2. All UI changes should be gated behind `QURAN_SYNC`. We need to be able to ship to the App Store without any sync-related code enabled. This also applies to localization too, create new strings rather than modify existing localizations

3. We shouldn't introduce new data-layer logic for new features (e.g., highlight) here. Most of that logic should exist in Kotlin. Adding data-layer logic makes sense only when adapting existing features to current protocols, not for new functionality like ayah collections

4. As discussed before, let's avoid adding migration code at this stage. Migration should come last, once everything is fully landed and we've aligned on the final UX and data impl.

AhmedNMahran:
> Thanks a lot for the PR. A couple of general comments:
> 
> 1. I'm seeing UI changes in this PR, even though it's described as data layer changes only PR
> 2. All UI changes should be gated behind `QURAN_SYNC`. We need to be able to ship to the App Store without any sync-related code enabled. This also applies to localization too, create new strings rather than modify existing localizations
> 3. We shouldn't introduce new data-layer logic for new features (e.g., highlight) here. Most of that logic should exist in Kotlin. Adding data-layer logic makes sense only when adapting existing features to current protocols, not for new functionality like ayah collections
> 4. As discussed before, let's avoid adding migration code at this stage. Migration should come last, once everything is fully landed and we've aligned on the final UX and data impl.

Thanks @mohamede1945 for the comments.
The points you mentioned are fair, I'll check the code again, regarding the collection names I need to check with BE the names whether they come directly from BE or I set locally and it appear there once synced.

Features/AppDependencies/AppDependencies.swift
mohamede1945, line 1: 1. I don't think we should have the new highlights code in the dependency injection file.
2. I also don't think we should mix the existing notes and highlights data service with the new ayah collection and bookmarking.

What I would do is:

1. Create a BookmarkCollection service that does CRUD operation on collections and ayah bookmarking. This shouldn't know anything about highlights or colors.
2. Add some helper methods for retreiving the colors collections to be used by the UI.
3. Integrate BookmarkCollection service into the UI with `#if QURAN_SYNC` checks. 

We can do something similar for the "new" Notes service. Probably name it differently like NotesSyncService or something and that shouldn't have anything to do with the existing NotesService powered by CoreData.

mohamede1945 (changes requested):

Domain/AnnotationsService/Sources/BookmarkCollectionService.swift
```diff
@@ -0,0 +1,90 @@
...
+                        for try await collections in syncService.collectionsWithBookmarksSequence() {
+                            continuation.yield(collections)
+                        }
+                        continuation.finish()
+                    } catch {
+                        continuation.finish()
+                    }
+                }
+                continuation.onTermination = { _ in task.cancel() }
+            }
```
mohamede1945: I think we can simply do `syncService.collectionsWithBookmarksSequence`

Domain/AnnotationsService/Sources/BookmarkCollectionService.swift
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        public func createCollection(named name: String) async throws {
+            try await syncService.createCollection(named: name)
+        }
+
+        public func addBookmark(sura: Int, ayah: Int, toCollectionLocalId localId: String) async throws {
```
mohamede1945: Instead of separate sura and ayah, let's use AyahNumber please.

Domain/AnnotationsService/Sources/BookmarkCollectionService.swift
```diff
@@ -0,0 +1,90 @@
...
+
+            for case let bookmark as CollectionBookmark.AyahBookmark in collection.bookmarks
+                where Int(bookmark.sura) == sura && Int(bookmark.ayah) == ayah
+            {
+                try await syncService.removeBookmarkFromCollection(
+                    collectionLocalId: localId,
+                    bookmark: bookmark.bookmark
+                )
+            }
+        }
```
mohamede1945: Let's not have this custom logic. UI should have a reference to a bookmark that can be passed and deleted directly. 

We can then remove snapshot() and BookmarkCollectionError unused code. Making BookmarkCollectionService just a wrapper for collections data model from mobile sync

Domain/AnnotationsService/Sources/NoteService.swift
```diff
@@ -57,6 +57,10 @@ public struct NoteService {
         _ = try await persistence.removeNotes(with: verses)
     }
 
+    public func removeHighlights(with verses: [AyahNumber]) async throws {
+        try await removeNotes(with: verses)
+    }
```
mohamede1945: I don't think we need this

Domain/AnnotationsService/Sources/QuranHighlightsSyncService.swift
```diff
@@ -0,0 +1,156 @@
+//
+//  QuranHighlightsSyncService.swift
```
mohamede1945: I don't think we need this service at all as I mentioned in my previous review. Please let me know if my reasoning is not clear.

Basically, highlights is just ayah bookmarking to a special named collections. There is no service or data layer logic to highlighting at all. It's a UI presentation thing. The whole concept of highlighting probably shouldn't be part of any services.

The only thing that I would extract out into a utility is the color mapping that's the collection name into a color is probably worth having it as a utility somewhere to be used by different UI

Features/AyahMenuFeature/AyahMenuBuilder+Sync.swift
mohamede1945, line 1: nit: Doesn't seem we need this file, the code can be inlined like we have for the non-sync situation or placed in the AyahMenuBuilder.swift file directly

Features/QuranContentFeature/ContentBuilder+Sync.swift
```diff
@@ -0,0 +1,31 @@
...
+
+import AnnotationsService
+import QuranImageFeature
+import QuranKit
+import QuranTranslationFeature
+import ReadingService
+
+extension ContentBuilder {
+    #if QURAN_SYNC
+        func makeSyncInteractorDeps(
```
mohamede1945: Same here, we probably either inline the code where it's used or put that code in ContentBuilder.swift file directly

Features/QuranContentFeature/ContentViewController.swift
```diff
@@ -86,6 +93,18 @@ public final class ContentViewController: UIViewController, UIGestureRecognizerD
...
+        #if QURAN_SYNC
+            syncHighlightsTask?.cancel()
+            syncHighlightsTask = Task { [weak self] in
+                guard let self else {
+                    return
+                }
+                await viewModel.observeSyncHighlightsIfNeeded()
+            }
+        #endif
+    }
```
mohamede1945: Let's not do this here, let's do this in the PagesView this way we don't have to manage the Task and the SwiftUI will manage it for us

Features/QuranViewFeature/QuranBuilder+Sync.swift
```diff
@@ -0,0 +1,43 @@
...
+import QuranContentFeature
+import QuranKit
+import ReadingService
+import TranslationsFeature
+import TranslationVerseFeature
+import WordPointerFeature
+
+extension QuranBuilder {
+    #if QURAN_SYNC
+        func makeSyncInteractorDeps(
```
mohamede1945: Same here too please inline or move it to the original file

Features/QuranViewFeature/QuranInteractor.swift
```diff
@@ -364,7 +367,14 @@ final class QuranInteractor: WordPointerListener, ContentListener, NoteEditorLis
...
         do {
-            try await deps.noteService.removeNotes(with: verses)
+            if notes.contains(where: { !($0.note ?? "").isEmpty }) {
+                try await deps.noteService.removeNotes(with: verses)
+            } else {
+                try await deps.noteService.removeHighlights(with: verses)
+                #if QURAN_SYNC
+                    try? await deps.highlightsSyncService?.removeHighlights(verses: verses)
+                #endif
+            }
```
mohamede1945: I don't think we should do this.

1. First this is UI thing and not data changes, there are few UI changes across the PR too
2. This is mixing notes and highlights, this is where I mentioned we need to really differentiate between them deleting a note in the old code meant delete note and associated highlights while in the new world, deleting a note doesn't necessitate we would delete a highlight. So basically we need different callbacks for these, like when we are in the old legacy core data, we have a deleteNoteAndHighlight and in the new world we have deleteNote and deleteHighlight so they are basically separate buttons doing separate actions
3. I don't think we should check whether the note exists in the core data stack to then decide we need to delete from the sync code

AhmedNMahran:
@mohamede1945 
I restacked this PR to keep it data-only.

What remains here is only the thin sync adapter layer:
- `MobileSyncSupport`
- `BookmarkCollectionService`
- `QURAN_SYNC` wiring in dependencies/container

I also moved the sync wrapper out of older modules so the new sync world is isolated instead of being folded into existing annotation/domain code.

All highlights UI/behavior was moved out to the next PR.

Example/QuranEngineApp/Classes/Container.swift
```diff
@@ -47,6 +47,10 @@ class Container: AppDependencies {
     }()
 
     private(set) lazy var notePersistence: NotePersistence = CoreDataNotePersistence(stack: coreDataStack)
+    #if QURAN_SYNC
+        private(set) lazy var syncService: SyncService? = mobileSyncServices?.syncService
+    #endif
+
```
mohamede1945, line 53: unused

mohamede1945 (approved):

PR #790, opened 2026-04-09
Add synced highlights UI

AhmedNMahran:
## Summary
Add sync-enabled highlights behavior and UI on top of the bookmark collection wrapper.

## Scope
- add Bookmarks `Highlights` row
- add fixed color list and per-color detail screen
- add sync-enabled highlight actions from the ayah menu
- render synced highlights on Quran content
- add separate synced ayah-menu flow instead of editing legacy note/highlight behavior

## Out of scope
- no notes sync
- no note editor changes
- no legacy non-sync behavior changes

## Notes
Sync-enabled highlights are based on fixed Mobile Sync collection names:
`red`, `green`, `blue`, `yellow`, `purple`.

Legacy non-sync behavior stays on the old path.

### DIFF
--- a/Features/AyahMenuFeature/SyncedAyahMenuBuilder.swift
+++ b/Features/AyahMenuFeature/SyncedAyahMenuBuilder.swift
@@ -1,0 +1,75 @@
+#if QURAN_SYNC
+    import AppDependencies
+    import FeaturesSupport
+    import MobileSync
+    import MobileSyncSupport
+    import QuranAnnotations
+    import QuranKit
+    import QuranTextKit
+    import UIKit
+
+    public struct SyncedAyahMenuInput {
+        // MARK: Lifecycle
+
+        public init(
+            sourceView: UIView,
+            pointInView: CGPoint,
+            verses: [AyahNumber],
+            notes: [QuranAnnotations.Note],
+            syncHighlightColor: HighlightColor?,
+            hasSyncHighlight: Bool
+        ) {
+            self.sourceView = sourceView
+            self.pointInView = pointInView
+            self.verses = verses
+            self.notes = notes
+            self.syncHighlightColor = syncHighlightColor
+            self.hasSyncHighlight = hasSyncHighlight
+        }
+
+        // MARK: Internal
+
+        let sourceView: UIView
+        let pointInView: CGPoint
+        let verses: [AyahNumber]
+        let notes: [QuranAnnotations.Note]
+        let syncHighlightColor: HighlightColor?
+        let hasSyncHighlight: Bool
+    }
+
+    @MainActor
+    public struct SyncedAyahMenuBuilder {
+        // MARK: Lifecycle
+
+        public init(container: AppDependencies) {
+            self.container = container
+        }
+
+        // MARK: Public
+
+        public func build(withListener listener: SyncedAyahMenuListener, input: SyncedAyahMenuInput) -> UIViewController {
+            let textRetriever = ShareableVerseTextRetriever(
+                databasesURL: container.databasesURL,
+                quranFileURL: container.quranUthmaniV2Database
+            )
+            let deps = SyncedAyahMenuViewModel.Deps(
+                sourceView: input.sourceView,
+                pointInView: input.pointInView,
+                verses: input.verses,
+                notes: input.notes,
+                syncHighlightColor: input.syncHighlightColor,
+                hasSyncHighlight: input.hasSyncHighlight,
+                syncService: container.syncService,
+                bookmarkCollectionService: container.bookmarkCollectionService,
+                textRetriever: textRetriever
+            )
+            let viewModel = SyncedAyahMenuViewModel(deps: deps)
+            viewModel.listener = listener
+            return SyncedAyahMenuViewController(viewModel: viewModel)
+        }
+
+        // MARK: Private
+
+        private let container: AppDependencies
+    }
+#endif
--- a/Features/BookmarksFeature/Sources/BookmarksBuilder.swift
+++ b/Features/BookmarksFeature/Sources/BookmarksBuilder.swift
@@ -26,10 +26,12 @@ public struct BookmarksBuilder {
         let viewModel = BookmarksViewModel(
             analytics: container.analytics,
             service: service,
+            highlightCollectionsUpdates: makeHighlightCollectionsUpdates(),
             authenticationClient: container.authenticationClient,
             navigateTo: { [weak listener] page in
                 listener?.navigateTo(page: page, lastPage: nil, highlightingSearchAyah: nil)
-            }
+            },
+            makeHighlightsController: makeHighlightsController(listener: listener)
         )
         let viewController = BookmarksViewController(viewModel: viewModel)
         return viewController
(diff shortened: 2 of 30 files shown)


AhmedNMahran:
@mohamede1945 
I split highlights behavior/UI out of the foundation PR.

In this PR I also stopped reusing the legacy note/highlight path for sync mode:
- sync-enabled highlight actions use a separate synced ayah-menu flow
- sync highlight rendering is separate from legacy note-backed highlights
- legacy non-sync behavior remains on the old path

The collection-to-highlight mapping is based on the fixed Mobile Sync collection names only.

Note:
Some duplication here is intentional: a small amount of existing feature-wiring logic was copied into a separate sync-enabled path so the new highlights flow stays isolated from the legacy note-backed flow. That keeps review clearer now and makes future deprecation of the old path easier.

Features/AyahMenuFeature/SyncedAyahMenuBuilder.swift
mohamede1945, line 1: There's a lot of duplication here (the whole SyncedAyahMenuXYZ) on top of the existing `AyahMenu` module. I think we should stick to our current approach: extend the existing code and gate changes behind `#if QURAN_SYNC`.

Reasons:

1. This level of duplication increases the risk of regressions in unrelated areas (e.g. playback, sharing).
2. It makes it harder to isolate and reason about the actual changes. We’re not fundamentally rewriting the module—we’re evolving it (splitting Notes/Highlights into Notes + Bookmark Collections with optional colors). Keeping changes localized will make future fixes and iterations significantly easier.
3. In fact, I'm not able to review these changes because I don't know what is new and what is old and I don't think I can do it easily manually.

Features/BookmarksFeature/Sources/BookmarksBuilder.swift
```diff
@@ -38,4 +40,34 @@ public struct BookmarksBuilder {
...
+                return nil
+            }
+
+            return {
+                HighlightCollection.updates(from: syncService)
+            }
+        #else
+            return nil
+        #endif
+    }
```
mohamede1945, line 72: We shouldn't have business logic here in DI layer. Whatever needs updates should get the service and call the method

Features/BookmarksFeature/Sources/BookmarksView.swift
```diff
@@ -68,23 +81,43 @@ private struct BookmarksViewUI: View {
             }
         }
         .task { await start() }
+        .task { await observeHighlights() }
```
mohamede1945, line 84: should simply be part of the start()

Features/BookmarksFeature/Sources/BookmarksView.swift
```diff
@@ -103,6 +136,24 @@ private struct BookmarksViewUI: View {
         )
     }
 
+    private var showsTopSections: Bool {
+        #if QURAN_SYNC
+            shouldShowSyncBanner || shouldShowHighlights
+        #else
+            shouldShowHighlights
+        #endif
```
mohamede1945, line 144: I don't think we need #if/else, it's embedded inside each of the props

UI/NoorUI/Components/List/NoorListItem.swift
```diff
@@ -44,6 +44,7 @@ public struct NoorListItem: View {
 
     public init(
         leadingEdgeLineColor: Color? = nil,
+        leadingView: AnyView? = nil,
```
mohamede1945, line 47: Instead of using AnyView in the API, we can do the following:

Init with a generic parameter
```
public init<LeadingView: View>
```

then parameter:
```
@ViewBuilder leadingView: () -> LeadingView? = { () -> EmptyView? in nil },
```

then store it as AnyView:
```
self.leadingView = leadingView().map { AnyView($0) }
```

This makes the API generic while we use AnyView in the impl. I'll later refactor it to remove the use of AnyView completely.

Features/BookmarksFeature/Sources/BookmarksView.swift
```diff
@@ -103,6 +136,24 @@ private struct BookmarksViewUI: View {
...
+    }
+
+    private var highlightsItem: some View {
+        NoorListItem(
+            leadingView: AnyView(HighlightPaletteIcon()),
```
mohamede1945, line 149: Remove the use of AnyView, see comment below in the NoorListItem for context, please

Features/FeaturesSupport/HighlightCollection.swift
```diff
@@ -0,0 +1,308 @@
...
+    public init(name: String, bookmarks: [HighlightBookmarkSnapshot]) {
+        self.name = name
+        self.bookmarks = bookmarks
+    }
+
+    public let name: String
+    public let bookmarks: [HighlightBookmarkSnapshot]
+}
+
+public enum HighlightCollection: Int, CaseIterable, Identifiable, Sendable {
```
mohamede1945, line 32: I don't think we need both HighlightCollection and HighlightColor

Features/FeaturesSupport/HighlightCollection.swift
```diff
@@ -0,0 +1,308 @@
...
+    public init(sura: Int, ayah: Int, modifiedDate: Date) {
+        self.sura = sura
+        self.ayah = ayah
+        self.modifiedDate = modifiedDate
+    }
+
+    public let sura: Int
+    public let ayah: Int
+    public let modifiedDate: Date
+}
```
mohamede1945, line 20: I don't think we need this type. This is mirroring: CollectionBookmark.AyahBookmark

Features/FeaturesSupport/HighlightCollection.swift
```diff
@@ -0,0 +1,308 @@
...
+
+public struct HighlightCollectionSnapshot: Equatable, Sendable {
```
mohamede1945, line 22: I believe this is trying to mirror CollectionWithBookmarks which is good. But few things:

1. This needs to not be highlight specific. I would name it `AyahBookmarksCollection`
2. This needs to exist in the data layer and have the service produce it instead of features doing so

mohamede1945 (commented):
Would you please provide screenshots of the changes?

Thanks!

Features/FeaturesSupport/HighlightCollection.swift
mohamede1945: Talked offline, we can do this in the mobile-sync

Features/FeaturesSupport/HighlightCollection.swift
```diff
@@ -0,0 +1,308 @@
...
+                    } catch {
+                        continuation.finish(throwing: error)
+                    }
+                }
+
+                continuation.onTermination = { _ in
+                    task.cancel()
+                }
+            }
+        }
```
mohamede1945, line 146: This method does nothing

Features/FeaturesSupport/HighlightCollection.swift
```diff
@@ -0,0 +1,308 @@
...
+                }
+                colorsByVerse[ayah] = color
+            }
+        }
+        return colorsByVerse
+    }
+
+    public func count(in collections: [HighlightCollectionSnapshot]) -> Int {
+        bookmarks(in: collections).count
+    }
```
mohamede1945, line 115: nit: I don't think we need this method

PR #793, opened 2026-04-23
Add notes sync data service

AhmedNMahran:
## Summary
Add the sync notes data layer only.

## Scope
- bump `mobile-sync-spm` to `0.0.7`
- add `NotesSyncService` in `MobileSyncSupport`
- wire the service through `AppDependencies` under `QURAN_SYNC`

## Behavior
- create always creates a sync note
- update/delete are by concrete sync note id

## Out of scope
- no notes UI
- no note editor changes
- no ayah menu icon/presentation changes
- no note tinting changes

## Notes
Legacy CoreData notes flow remains untouched in this PR.

### DIFF
--- a/Data/MobileSyncSupport/Sources/NotesSyncService.swift
+++ b/Data/MobileSyncSupport/Sources/NotesSyncService.swift
@@ -66,6 +66,10 @@
             Int64(QuranData.shared.getAyahId(sura: Int32(verse.sura.suraNumber), ayah: Int32(verse.ayah)))
         }
 
+        // TODO: Remove/Revisit this workaround once Mobile Sync stops throwing
+        // "Expected note after insert." for notes that were actually persisted.
+        // Today note creation can succeed in the sync store but still surface that
+        // false-negative error, so we verify the inserted note exists before failing.
         private static func isExpectedNoteAfterInsert(_ error: Error) -> Bool {
             let description = String(describing: error)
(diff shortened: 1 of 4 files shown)


AhmedNMahran:
@mohamede1945 This PR is now data-only.

It only adds the sync notes service and dependency wiring. Notes UI/editor behavior was moved out to the follow-up PR so data and presentation are reviewed separately.

Data/MobileSyncSupport/Sources/NotesSyncService.swift
```diff
@@ -0,0 +1,68 @@
...
+            self.syncService = syncService
+        }
+
+        // MARK: Public
+
+        public func createNote(body: String, verses: Set<AyahNumber>) async throws {
+            guard let range = Self.ayahIdRange(for: verses) else {
+                return
+            }
+            try await syncService.createNote(body: body, startAyahId: range.start, endAyahId: range.end)
```
mohamede1945: Given that a note can be added to a continuous range of ayat, we should make the method accept start ayah and end ayah, basically matching the SyncService API. The current parameter can be a set of disjointed ayat which then makes the persistence incorrect.

AhmedNMahran: @mohamede1945 Done, check again please

PR #795, opened 2026-04-25
Add navigation drawer to reading screen #794

### LINKED ISSUE #794: Feature: in-reading navigation drawer for surahs / juz / notes / bookmarks
## Navigation Drawer — quick in-reading access

A side-sheet that slides in from the trailing edge of the reading screen with one-tap access to:
                                                                                
- **Surah list** — jump to any of the 114 surahs                             
- **Juz list** — 240 quarters grouped by juz, like the Home tab               
- **Notes** — your saved notes, tap to jump to the verse                                                 
- **Bookmarks** — your bookmarked pages, tap to jump                          
- **Search** — filters within the active section                                                         
                                                                                
**Why it matters:** today, switching to a different surah or reviewing a note means backing out of the   
reading view to the main tabs, then re-entering — breaking your reading flow. The drawer keeps you in
context: tap the title bar → pick a destination → drawer slides closed → you're there. Especially        
valuable during long reading sessions or when cross-referencing.

abdorizak:
Closes #794                                                                                              
                                                                                                           
  ## What it does                                                                                          
                                                                                                           
  Adds a side-sheet you open by **tapping the title in the reading nav bar**. From there you can jump to   
  any:                                                                                                     
                                        
  - **Surah** (114, grouped by juz)                                                                        
  - **Juz / quarter** (240 quarters, same layout as Home)                                                  
  - **Note** (your saved notes)                                                
  - **Bookmark** (your bookmarked pages)                                                                   
                                                                                                           
  …with a search field that filters the active tab.                             
                                                                                                           
  ## Why                                                          
                                                                                                           
  Today, switching surahs or checking a note means backing out of the reading view to the main tabs and    
  re-entering — interrupting the reading flow. The drawer keeps you in context: tap title → pick           
  destination → keep reading.                                                                              
                                                                  
  ## How                                                                                                   
   
  - New `NavigationDrawerFeature` SPM target. Drawer is built only when the title is tapped — nothing new  
  runs on the reading / scrolling path.                                         
  - Reuses `NoorList` / `NoorListItem` so rows look the same as the Home tab.
  - Strings added to all 16 locales (`en`/`ar` hand-written, the rest AI-translated and labelled per the
  repo's existing convention).                                                                             
                                                                                
  ## Demo                                                                                                  
                                                                                                           
  > Uploading iPhone 16E Screen Recording Apr 25 2026.mp4…

                                                                  
                                                                                                           
  ## Tested                                                                                                
                                                                                                           
  - First-tap from cold start, recents, surah list, and juz list → drawer opens reliably
  - Search filters each tab                                                                                
  - Selecting a destination navigates and dismisses the drawer                  
  - RTL + dark mode both look right                                                                        
                                                                               
  Happy to address feedback. 🙏

### DIFF
--- a/Core/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Resources/ar.lproj/Localizable.strings
@@ -224,3 +224,22 @@
 "new.library_navigation" = "التنقل:";
 // Metadata: Translated by ChatGPT; human review required.
 "new.library_navigation.details" = "* قم بترتيب السور والأجزاء على الصفحة الرئيسية، وعرض أرقام السور حتى تتمكن من الانتقال إلى ما تحتاجه بشكل أسرع - بفضل @yismailuofa.";
+
+// MARK: - Navigation Drawer
+// Metadata: Translated by AI; human review required.
+"navigation_drawer.title" = "التنقل";
+"navigation_drawer.search_placeholder" = "بحث";
+"navigation_drawer.tab.surah" = "السور";
+"navigation_drawer.tab.juz" = "الأجزاء";
+"navigation_drawer.tab.notes" = "الملاحظات";
+"navigation_drawer.tab.bookmarks" = "الإشارات المرجعية";
+"navigation_drawer.surah.makki" = "مكية";
+"navigation_drawer.surah.madani" = "مدنية";
+"navigation_drawer.note.untitled" = "ملاحظة بدون عنوان";
+"navigation_drawer.bookmark.page" = "صفحة %d";
+"navigation_drawer.empty.no_results" = "لا توجد نتائج";
+"navigation_drawer.empty.no_notes" = "لا توجد ملاحظات بعد";
+"navigation_drawer.empty.no_matching_notes" = "لا توجد ملاحظات مطابقة";
+"navigation_drawer.empty.no_bookmarks" = "لا توجد إشارات مرجعية بعد";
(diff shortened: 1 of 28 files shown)


abdorizak:
@mohamede1945 Please check this thanks.

mohamede1945:
Jazak Allah khayran for the implementation, this looks really well done.

That said, I don't think we can move forward with this change right now for a few reasons:

1. We're currently working on a broader redesign of the app, which will likely impact how this feature should be introduced.
2. The current interaction isn't very discoverable, so it may not deliver the intended value to users.
3. It modifies the existing tap behavior in the reading view. While the approach itself is reasonable, given the upcoming redesign, it's probably better to avoid introducing a new interaction pattern that we may soon change.

I do agree that this is a valuable feature and something we should support. It just feels like the timing and UX approach need to align better with the upcoming redesign.

Appreciate the work here. I would love to revisit this once we have more clarity on the new direction.

abdorizak:
Thanks for the detailed feedback, that makes sense.

I agree on the timing and the discoverability concern. Happy to pause on this for now and revisit once the redesign direction is clearer. I still think improving in-context navigation during reading is valuable, so would love to pick this back up when it aligns better.

In the meantime, I’m happy to help with other feature work or contribute to the new design efforts. just let me know where I can be most useful.

Appreciate the guidance. @mohamede1945

mohamede1945:
@abdorizak I would appreciate you taking a look at https://github.com/quran/quran-ios/issues/809. There are other tickets in the repo, you could take a look, but let me know by commenting first on the issue since some are out dated

PR #801, opened 2026-04-26
Add new data fields to support Naskh

ahmedre:
Add a mapping table for juz' to page number, a total page count, and a
number of pages to "skip" from the start.

### DIFF
--- a/Model/QuranKit/Sources/Quran.swift
+++ b/Model/QuranKit/Sources/Quran.swift
@@ -29,6 +29,14 @@ public final class Quran: Hashable, @unchecked Sendable {
         raw.arabicBesmAllah
     }
 
+    public var numberOfPages: Int {
+        raw.numberOfPages
+    }
+
+    public var pagesToSkip: Int {
+        raw.pagesToSkip
+    }
+
     public var suras: [Sura] {
         lazySuras()
     }
@@ -78,7 +86,7 @@ public final class Quran: Hashable, @unchecked Sendable {
 
 extension Quran {
     var pagesRange: ClosedRange<Int> {
-        1 ... raw.startSuraOfPage.count
+        (raw.pagesToSkip + 1) ... raw.numberOfPages
     }
 }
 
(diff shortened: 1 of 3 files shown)


ahmedre:
we need this because the first page is numbered "page 2" in the mushaf, but it's on the right side. consequently, we add a "skip pages" variable on Android to skip the first page and to "hack" odd/even page numbering.

Model/QuranKit/Sources/ReadingInfo/Madani1405QuranReadingInfoRawData.swift
```diff
@@ -23,6 +23,14 @@ struct Madani1405QuranReadingInfoRawData: QuranReadingInfoRawData {
...
     }
 
+    var startPageOfJuz: [Int] {
+        [
+            1, 22, 42, 62, 82, 102, 121, 142, 162, 182,
+            201, 222, 242, 262, 282, 302, 322, 342, 362, 382,
+            402, 422, 442, 462, 482, 502, 522, 542, 562, 582,
+        ]
+    }
+
```
mohamede1945: I don't think we need this. We can get it from the page of the first verse of the juz. Could you explain this one please?

Model/QuranKit/Sources/Quran.swift
```diff
@@ -77,8 +85,8 @@ public final class Quran: Hashable, @unchecked Sendable {
 }
 
 extension Quran {
-    var pagesRange: ClosedRange<Int> {
-        1 ... raw.startSuraOfPage.count
+    var pagesRange: Range<Int> {
```
mohamede1945: Is there a reason why we have this as a Range rather than ClosedRange?

Model/QuranKit/Sources/ReadingInfo/Madani1405QuranReadingInfoRawData.swift
ahmedre: I was worried initially that because these page numbers are different for Naskh than they are for Madani that we'd need this array, but agree you are right - will fix in sha' Allah.

Model/QuranKit/Sources/Quran.swift
ahmedre: no good reason - changing it back

mohamede1945 (approved):

PR #803, opened 2026-04-26
Add a QuranPageMapper for mapping bookmark types

ahmedre:


ahmedre:
this isn't connected until the next pr, but the point of this pr is to add a mapper between Quran page types. for example, Naskh is 611 pages, and normal madani is 604 pages.

what it does is as follows:
for page bookmarks, gets the first ayah according to the current page type, and uses that to figure out the page in the target Quran page type. for ayah bookmarks, it just updates the page accordingly.

the next pr is the one that connects it.

Model/QuranKit/Sources/QuranPageMapper.swift
```diff
@@ -0,0 +1,65 @@
...
+//  QuranPageMapper.swift
+//
+//
+//  Created by OpenAI on 2026-04-25.
+//
+
+public struct QuranPageMapper {
+    // MARK: Lifecycle
+
+    public init(source: Quran, destination: Quran) {
```
mohamede1945: I'd prefer if we always use `Page` and `AyahNumber` rather than `Int`s. That means we can just have destination and source will be from the input.

mohamede1945 (approved):

Model/QuranKit/Sources/QuranPageMapper.swift
```diff
@@ -0,0 +1,65 @@
...
+        return mapAyah(sourcePage.firstVerse)
+    }
+
+    public func mapAyah(_ ayah: AyahNumber) -> Page? {
+        mapAyah(sura: ayah.sura.suraNumber, ayah: ayah.ayah)
+    }
+
+    public func mapAyah(sura: Int, ayah: Int) -> Page? {
+        AyahNumber(quran: destination, sura: sura, ayah: ayah)?.page
+    }
```
mohamede1945: This is probably can return AyahNumber directly and consumers can use the page if needed

Model/QuranKit/Sources/QuranPageMapper.swift
ahmedre: done

Model/QuranKit/Sources/QuranPageMapper.swift
ahmedre: done

PR #804, opened 2026-04-26
Transparently map verses when talking to sync

ahmedre:
Quran.com's sync only deals with Madani mushaf numbering. For naskh and
others, transparently map between the ui and the database, so that
syncing and the source of truth are madani numbering, but the actual
pages can use different numbering.

### DIFF
--- a/Domain/AnnotationsService/Sources/LastPageService.swift
+++ b/Domain/AnnotationsService/Sources/LastPageService.swift
@@ -14,42 +14,78 @@ import QuranKit
 public struct LastPageService {
     // MARK: Lifecycle
 
-    public init(persistence: LastPagePersistence) {
+    public init(persistence: LastPagePersistence, storedPageQuran: Quran = .hafsMadani1405) {
         self.persistence = persistence
+        self.storedPageQuran = storedPageQuran
     }
 
     // MARK: Public
 
     public func lastPages(quran: Quran) -> AnyPublisher<[LastPage], Never> {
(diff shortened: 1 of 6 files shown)


ahmedre:
I made some intentional decisions worth calling out here:
1. from the QuranSync perspective, _everything_ is Madani 1405 mapped (since today, we have page type in the api, but only Madani 1405 is supported).
2. mapping is transparent between the data layer / sync layer and the ui layer. consequently, the data and sync layers are always 1405 based.
3. I thought about making a "no op" flavor of this to not do any mapping, but there are some cases, even between the madani 1405 style data and the newer madani data. for example, here's a map of some of the page start changes between the two masahif -

```
121 5:77->5:78, 122 5:83->5:84, 123 5:90->5:91, 145 6:132->6:131,
532 55:17->55:19, 533 55:41->55:42, 534 55:68->55:70, 565 68:16->68:17, 568 69:35->69:36, 
570 70:40->70:41, 576 74:18->74:19, 584 79:16->79:17, 586 81:1->80:41, 588 83:7->83:5,
589 83:35->83:34, 590 85:1->84:25, 592 87:16->87:11, 593 89:1->88:23, 594 89:24->89:23,
595 91:1->90:19, 596 92:15->92:10, 597 95:1->94:3, 598 97:1->96:13, 599 98:8->98:6,
600 100:10->100:6
```

cc @AhmedNMahran

mohamede1945 (approved):

PR #805, opened 2026-04-26
Parameterize line pages

ahmedre:
Madani line pages (1439, 1441) render with slightly different
parameters, line ratios, etc to naskh. Moreover, madani lines are made
to slightly overlap each other safely, whereas this is not the case with
naskh lines, which need to split the spacing equally. This patch makes
these parameters.


ahmedre:
screenshot within the last pr

mohamede1945 (approved):

PR #808, opened 2026-04-26
Add and wire up naskh page type

ahmedre:


ahmedre:
full page
Simulator Screenshot - iPhone 17 Pro - 2026-04-26 at 21 38 00

with sidelines
Simulator Screenshot - iPhone 17 Pro - 2026-04-26 at 21 37 51

with sidelines
Simulator Screenshot - iPhone 17 Pro - 2026-04-26 at 21 37 23

note that when a page doesn't have sidelines it takes up all available space as in the first example

ahmedre:
things left that we should (ideally) do before shipping:
1. add a setting for enabling / disabling sidelines
2. add a setting for enabling / disabling lines between the lines
3. when choosing naskh pages, change to the naskh font and text for translation mode
4. when choosing naskh pages, show the manzil number next to the juz' / hizb number
5. in naskh pages mode, show naskh naming of suras instead of ordinary madani ones (ex sura Dahr instead of sura Insan, etc).

mohamede1945 (approved):

PR #810, opened 2026-04-28
Feature/collapsible juz sections

### LINKED ISSUE #800: Collapsible Juz's sections
From email:

> I have a suggestion for the Juz’ view: the list is very long. For example, if I am in Al-A‘raf, Juz’ 9, the first half (Hizb 17), I have to scroll down a lot to reach it. If possible, could you please add a minimize/collapse button for each Juz’? It would make it much easier to navigate to the desired Hizb. JazakAllahu khayran.

OKKHALIL3:
Assalamu alaykum,

  Small contribution to address #800 — long Juz' lists in the Home view.

  ## Changes
  - Each Juz' section header is now tappable with a chevron to collapse/expand its rows.
  - Added a navigation bar button to collapse/expand all Juz' sections at once. Icon and accessibility label
  switch based on current state.
  - Applies to both the Suras and Juzs tabs since both group by Juz.

  State is in-memory only — happy to follow up with persistence across launches if maintainers want it.

  Closes #800


Features/HomeFeature/HomeViewModel.swift
```diff
@@ -56,12 +56,39 @@ final class HomeViewModel: ObservableObject {
 
     @Published var surahSortOrder: SurahSortOrder = HomePreferences.shared.surahSortOrder
 
+    @Published var collapsedJuzs: Set<Int> = []
```
mohamede1945: Let's use `Juz` instead of Int please for type safety. Same for the methods and callbacks please

UI/NoorUI/Components/List/NoorCollapsibleSection.swift
```diff
@@ -0,0 +1,66 @@
+//
+//  NoorCollapsibleSection.swift
+//
+//
+//  Created by QuranEngine on 2026.
+//
+
+import SwiftUI
+
+public struct NoorCollapsibleSection<Item: Identifiable, ListItem: View>: View {
```
mohamede1945: Instead of creating a new section, we should augment NoorSection with this functionality. 

We should also be using one of:
- https://developer.apple.com/documentation/swiftui/section/init(_:isexpanded:content:)
- https://developer.apple.com/documentation/swiftui/section/init(isexpanded:content:header:)

mohamede1945 (changes requested):
This is great, mashaa'Allah. Jazak Allah khyrn for implement it.

mohamede1945:
Oh one more thing I forgot to mention, for UI changes, please include screenshots or video recording of the changes.

Thank you!

OKKHALIL3:
Jazak Allah Khair for the review! I've pushed the requested changes:

1. **`Juz` instead of `Int`** — `collapsedJuzs` is now `Set<Juz>`, and `isJuzExpanded`/`setJuz` plus the `HomeView` callbacks all take `Juz` directly for type safety.
2. **Augmented `NoorSection`** — removed the separate `NoorCollapsibleSection` and added an optional `isExpanded: Binding<Bool>?` parameter to `NoorSection`. When provided, it renders the section as collapsible.

One note on the native API suggestion: I tried `Section(isExpanded:content:header:)` (iOS 17+) with an iOS 15/16 fallback, but it didn't render a visible disclosure chevron in `.insetGrouped` list style in my testing — the section just appeared without any affordance to collapse. So I kept the manual button + rotating chevron implementation inside `NoorSection` so the UI works consistently across the supported iOS versions. Happy to revisit if you've seen the native API render the chevron in this list style.

OKKHALIL3:
https://github.com/user-attachments/assets/61f7d671-07be-4e9e-81cf-390ac2c7a769

Ive attached above a video of the ui.

mohamede1945:
Jazak Allah khyrn for the changes!

PR #813, opened 2026-04-29
Fix another flakey test

ahmedre:


Domain/ReadingService/Sources/ReadingResourceDownloader.swift
```diff
@@ -82,6 +82,6 @@ private extension RemoteResource {
     }
 
     func matches(_ request: DownloadRequest) -> Bool {
-        request.destination == downloadDestination
+        request.url == url && request.destination == zipFile
```
ahmedre: this makes me uneasy, but from my understanding this is correct -
`request.destination` is actually a file path (ex /path/to/tajweed.zip), whereas `downloadDestination` is a directory (ex /path/to/) - consequently the previous check would not work.

probably zipFile isn't the ideal name but in most cases it's a zip file so maybe it's ok?

ahmedre (commented):
honestly only got this flakey test once so am ok dropping this pr if we feel the risk is not worth it

Domain/ReadingService/Sources/ReadingResourceDownloader.swift
mohamede1945: Oh, the previous check was incorrect. Thanks for catching it. I checked the code and it happens in a very rare scenario, mostly when the app is terminated while the backend download is running and then user opens .

 I would skip `request.url == url` check since in some situations the request's url could change based on redirects, so to be in the safe side, let's just do the destination check

mohamede1945 (approved):

PR #820, opened 2026-05-04
Add sync service integration for reading sessions in Home and ContentView models

AhmedNMahran:


mohamede1945 (changes requested):
I think we need a `MobileSyncLastPageService` implementation of `LastPageService`.

We can turn `LastPageService` into a protocol and provide two implementations:

1. The existing implementation.
2. A new `MobileSyncLastPageService` implementation for mobile sync.

Thank you!

mohamede1945 (approved):

PR #821, opened 2026-05-05
Add collections feature to bookmarks, including UI and view model integration

AhmedNMahran:
- show collections item in bookmarks page
- each collection item is expandable to show it's bookmarks
- swipe to delete
- delete bookmark, collection

### Notes:
- still Not implemented add collection ayah bookmark from the app
- shows only bookmarks from the api
- highlights not yet supported

| Screenshots | Screenshots |
|--------|--------|
|  Screenshot |  Screenshot 2026-05-06 at 3 33 39 AM |
|  Screenshot 2026-05-06 at 3 37 44 AM | Screenshot |

### DIFF
--- a/Core/Localization/Resources/en.lproj/Localizable.strings
+++ b/Core/Localization/Resources/en.lproj/Localizable.strings
@@ -243,3 +243,10 @@
 // Library & Navigation
 "new.library_navigation" = "Library & Navigation:";
 "new.library_navigation.details" = "* Sort surahs and juz on Home, and display surah numbers so you can jump to what you need faster - thanks to @yismailuofa.";
+
+"bookmarks.collections" = "Collections";
+"bookmarks.collections.add" = "Add Collection";
+"bookmarks.collections.edit" = "Edit Collections";
+"bookmarks.collections.new.placeholder" = "Collection name";
+"bookmarks.collections.no-data.title" = "No collections yet";
+"bookmarks.collections.no-data.text" = "Create a collection to organize your ayah bookmarks.";
--- a/Core/Localization/Resources/en.lproj/Localizable.strings
+++ b/Core/Localization/Resources/en.lproj/Localizable.strings
@@ -250,3 +250,5 @@
 "bookmarks.collections.new.placeholder" = "Collection name";
 "bookmarks.collections.no-data.title" = "No collections yet";
 "bookmarks.collections.no-data.text" = "Create a collection to organize your ayah bookmarks.";
+
+"bookmarks.old-page-bookmarks" = "Old Page Bookmarks";
--- a/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
+++ b/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
@@ -84,9 +84,13 @@
         }
 
         public func collectionsSequence() -> AyahBookmarkCollectionsSequence {
+            let readingPreferences = readingPreferences
             let sequence = syncService.collectionsWithBookmarksSequence()
                 .map { collections in
-                    self.collections(from: collections)
+                    Self.collections(
+                        from: collections,
+                        quran: readingPreferences.reading.quran
+                    )
                 }
             return AyahBookmarkCollectionsSequence(sequence)
         }
@@ -117,10 +121,6 @@
             }
 
             return AyahCollectionBookmark(bookmark: bookmark, ayah: ayah)
-        }
-
-        private func collections(from collections: [CollectionWithAyahBookmarks]) -> [AyahBookmarkCollection] {
-            Self.collections(from: collections, quran: readingPreferences.reading.quran)
         }
     }
 
--- a/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
+++ b/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
@@ -32,13 +32,13 @@
 
         // MARK: Public
 
-        public func observeCollections(_ handler: ([AyahBookmarkCollection]) -> Void) async throws {
+        public func observeCollections(_ handler: @MainActor ([AyahBookmarkCollection]) -> Void) async throws {
             let sequence = syncService.collectionsWithBookmarksSequence()
                 .map { collections in
                     self.collections(from: collections)
                 }
             for try await collections in sequence {
-                handler(collections)
+                await handler(collections)
             }
         }
 
--- a/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
+++ b/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
@@ -75,10 +75,6 @@
         }
     }
 
-    // TODO: Move this extension to mobile-sync repository in a follow-up PR.
-    // This is a temporary workaround to support CollectionAyahBookmark directly.
-    // The mobile-sync API should be enhanced to accept CollectionAyahBookmark
-    // instead of requiring conversion to AyahBookmark.
     extension SyncService {
         func removeBookmarkFromCollection(_ bookmark: AyahCollectionBookmark) async throws {
             try await removeBookmarkFromCollection(
--- a/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
+++ b/Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
@@ -9,113 +9,106 @@
     import QuranKit
     import ReadingService
 
-    struct AyahBookmarkCollection {
-        let collection: Collection_
-        let bookmarks: [AyahCollectionBookmark]
+    public struct AyahBookmarkCollection {
+        public let collection: Collection_
+        public let bookmarks: [AyahCollectionBookmark]
     }
 
-    struct AyahCollectionBookmark {
-        let bookmark: CollectionAyahBookmark
-        let ayah: AyahNumber
+    public struct AyahCollectionBookmark {
+        public let bookmark: CollectionAyahBookmark
+        public let ayah: AyahNumber
     }
 
-    struct AyahBookmarkCollectionService {
+    public struct AyahBookmarkCollectionService {
         // MARK: Lifecycle
 
-        init(readingPreferences: ReadingPreferences = .shared) {
-            self.init(syncService: nil, readingPreferences: readingPreferences)
-        }
-
-        init(
-            syncService: SyncService?,
+        public init(
+            syncService: SyncService,
             readingPreferences: ReadingPreferences = .shared
         ) {
             self.syncService = syncService
             self.readingPreferences = readingPreferences
         }
 
-        // MARK: Internal
+        // MARK: Public
 
-        func collectionsSequence() -> AsyncThrowingStream<[AyahBookmarkCollection], Error> {
-            AsyncThrowingStream { continuation in
-                let task = Task {
-                    do {
-                        let syncService = try requireSyncService()
-                        for try await collections in syncService.collectionsWithBookmarksSequence() {
-                            continuation.yield(self.collections(from: collections))
-                        }
-                        continuation.finish()
-                    } catch {
-                        continuation.finish(throwing: error)
-                    }
+        public func observeCollections(_ handler: ([AyahBookmarkCollection]) -> Void) async throws {
+            let sequence = syncService.collectionsWithBookmarksSequence()
+                .map { collections in
+                    self.collections(from: collections)
                 }
-                continuation.onTermination = { _ in task.cancel() }
+            for try await collections in sequence {
+                handler(collections)
             }
         }
 
-        func createCollection(named name: String) async throws {
-            let syncService = try requireSyncService()
+        public func collectionsSnapshot() async throws -> [AyahBookmarkCollection] {
+            let sequence = syncService.collectionsWithBookmarksSequence()
+                .map { collections in
+                    self.collections(from: collections)
+                }
+            var iterator = sequence.makeAsyncIterator()
+            return try await iterator.next() ?? []
+        }
+
+        public func createCollection(named name: String) async throws {
             try await syncService.createCollection(named: name)
         }
 
-        func removeCollection(localId: String) async throws {
-            let syncService = try requireSyncService()
-            try await syncService.removeCollection(localId: localId)
-        }
-
-        // TODO: Move this conversion to mobile-sync-spm once it supports deleting
-        // collection bookmarks directly without converting to AyahBookmark first.
-        func removeBookmarkFromCollection(_ bookmark: AyahCollectionBookmark) async throws {
-            let syncService = try requireSyncService()
-            try await syncService.removeBookmarkFromCollection(
-                collectionLocalId: bookmark.bookmark.collectionLocalId,
-                bookmark: AyahBookmark(
-                    sura: bookmark.bookmark.sura,
-                    ayah: bookmark.bookmark.ayah,
-                    lastUpdated: bookmark.bookmark.lastUpdated,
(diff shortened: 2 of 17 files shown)


Core/Localization/Resources/en.lproj/Android.strings
```diff
@@ -185,6 +185,7 @@ sure you want to use the external SD card?";
 "menu_bookmarks" = "Bookmarks";
 "menu_bookmarks_ayah" = "Ayah Bookmarks";
 "menu_bookmarks_page" = "Page Bookmarks";
+"bookmarks.collections" = "Collections";
```
mohamede1945: Please don't add localization to the Android strings these are auto-imported from the Android repo

Features/BookmarksFeature/Sources/BookmarksBuilder.swift
```diff
@@ -32,6 +32,16 @@ public struct BookmarksBuilder {
...
+        #if QURAN_SYNC
+            let collectionsBuilder = CollectionsBuilder(container: container)
+            viewController.onOpenCollections = { [weak viewController, weak listener] in
+                guard let listener else {
+                    return
+                }
+                let collectionsViewController = collectionsBuilder.build(withListener: listener)
+                viewController?.navigationController?.pushViewController(collectionsViewController, animated: true)
+            }
+        #endif
```
mohamede1945, line 74: We don't want to put business logic in the builders. In this case the navigation logic should live inside and not be exposed

Features/BookmarksFeature/Sources/BookmarksViewController.swift
```diff
@@ -42,6 +46,14 @@ final class BookmarksViewController: UIHostingController<BookmarksView> {
...
         title = lAndroid("menu_bookmarks")
         viewModel.presenter = self
+        #if QURAN_SYNC
+            rootView = BookmarksView(
+                viewModel: viewModel,
+                showCollectionsAction: { [weak self] in
+                    self?.onOpenCollections?()
+                }
+            )
+        #endif
```
mohamede1945: I don't think we should have 2 places to init the rootView, this makes the logic harder to reason about.

Features/BookmarksFeature/Sources/CollectionsView.swift
```diff
@@ -0,0 +1,117 @@
...
+            return [
+                NoorSectionHeaderAction(image: .delete, tintColor: .red) {
+                    await deleteCollectionAction(collection)
+                },
+            ]
+        }
+
+        @ViewBuilder
+        private func bookmarkItem(_ bookmark: CollectionAyahBookmark) -> some View {
+            if let ayah = AyahNumber(quran: ReadingPreferences.shared.reading.quran, sura: Int(bookmark.sura), ayah: Int(bookmark.ayah)) {
```
mohamede1945: This is a business logic that shouldn't live here. 

What we need is a AyahBookmarkCollectionService type, that basically do this conversion and give us data types with AyahNumber.

Please create this service in a separate PR.

In fact I don't think we will be using the currently reading quran, but I think we can just add a TODO for it in the new service, until we figure out how we should handle different mushafs.

Features/BookmarksFeature/Sources/CollectionsViewController.swift
```diff
@@ -0,0 +1,85 @@
...
+                    },
+                    UIAction(
+                        title: viewModel.editMode.isEditing ? l("button.done") : l("bookmarks.collections.edit"),
+                        image: UIImage(systemName: viewModel.editMode.isEditing ? NoorSystemImage.checkmark.rawValue : NoorSystemImage.edit.rawValue)
+                    ) { [weak self] _ in
+                        self?.toggleEditing()
+                    },
+                ])
+            )
+        }
```
mohamede1945: When in editing mode, we shouldn't use a UIMenu, we should show a done button only for easy exit of the editing mode

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewController.swift
mohamede1945, line 1: I think this is missing a rename action

Features/BookmarksFeature/Sources/CollectionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+            if expanded {
+                collapsedCollectionIDs.remove(id)
+            } else {
+                collapsedCollectionIDs.insert(id)
+            }
+        }
+
+        func toggleCollection(_ collection: CollectionWithAyahBookmarks) {
+            setCollection(collection, expanded: !isCollectionExpanded(collection))
+        }
```
mohamede1945: Unused, please delete

Features/BookmarksFeature/Sources/CollectionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+
+        func navigateTo(_ bookmark: CollectionAyahBookmark) {
+            logger.info("Bookmarks: select collection bookmark at \(bookmark.sura):\(bookmark.ayah)")
+            guard let ayah = AyahNumber(quran: readingPreferences.reading.quran, sura: Int(bookmark.sura), ayah: Int(bookmark.ayah)) else {
+                return
+            }
```
mohamede1945: Same as above, this is business logic that should live in the AyahBookmarkCollectionService

Features/BookmarksFeature/Sources/CollectionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+                let ayahBookmark = AyahBookmark(
+                    sura: bookmark.sura,
+                    ayah: bookmark.ayah,
+                    lastUpdated: bookmark.lastUpdated,
+                    localId: bookmark.bookmarkLocalId
+                )
+                try await syncService.removeBookmarkFromCollection(
+                    collectionLocalId: bookmark.collectionLocalId,
+                    bookmark: ayahBookmark
+                )
```
mohamede1945: I think we need to add to mobile-sync a delete/remove method that takes CollectionAyahBookmark directly

Basically this file https://github.com/quran/mobile-sync/blob/5f8145fc6553c4c9cc9fbe099ecfc27d2d044653/persistence/src/commonMain/kotlin/com/quran/shared/persistence/repository/collectionbookmark/repository/CollectionBookmarksRepository.kt

works with CollectionAyahBookmark for all methods except the delete, so I think it makes more sense to have a delete of CollectionAyahBookmark

UI/NoorUI/Components/List/NoorSection.swift
```diff
@@ -116,22 +167,31 @@ public struct NoorSection<Item: Identifiable, ListItem: View>: View {
...
         self.onDelete = onDelete
         self.onMove = onMove
     }
 
     // MARK: Public
 
     public var body: some View {
-        if !items.isEmpty {
-            NoorBasicSection(title: title, isExpanded: isExpanded) {
+        if !items.isEmpty || title != nil {
```
mohamede1945: remove the title check please

Core/Localization/Resources/en.lproj/Localizable.strings
mohamede1945, line 1: Please also add translations to the Arabic file

mohamede1945 (changes requested):

Features/BookmarksFeature/Sources/CollectionsViewController.swift
AhmedNMahran: editing mode would include renaming, that's why I added a menu, but renaming is not exposed from the spm today, so to be added later insha'Allah, if you still prefer the done button way, I'll do it no issue insha'Allah

mohamede1945: Sorry, I'm not sure I understand how the menu would include renaming. This menu is for the entire list of collections. The renaming would be on individual collection items. Could you explain please?

The general principle here is that, when a user enters the editing mode (a non-default view), it should be easy and obvious for them to exit from that mode back to the default view.

Thank you!

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewController.swift
AhmedNMahran: to be added in a follow-up after exposing the function on the mobile-sync-spm insha'Allah

Features/BookmarksFeature/Sources/CollectionsViewController.swift
AhmedNMahran: oh sorry I misunderstood the first comment, sure that's what I did with delete, the rename should be added also in the non-default view after I expose the spm-function insha'Allah

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
mohamede1945, line 1: This is repeated from the other PR

Features/BookmarksFeature/Sources/BookmarksViewController.swift
```diff
@@ -58,6 +75,17 @@ final class BookmarksViewController: UIHostingController<BookmarksView> {
             .store(in: &cancellables)
     }
 
+    #if QURAN_SYNC
+        private func openCollections() {
+            guard let container, let listener else {
+                return
+            }
+            let collectionsBuilder = CollectionsBuilder(container: container)
```
mohamede1945: So we shouldn't pass `container` to modules this is a DI container and by doing so we are moving from DI to service registry pattern.

So what we need to do is create CollectionsBuilder in the builder and pass it instead of passing container

Features/BookmarksFeature/Sources/CollectionsView.swift
```diff
@@ -0,0 +1,137 @@
...
+                }
+            }
+            .task { await start() }
+            .errorAlert(error: $error)
+            .environment(\.editMode, $editMode)
+        }
+
+        // MARK: Private
+
+        private let ayahBookmarkCollectionService = AyahBookmarkCollectionService()
```
mohamede1945: This needs to move to the ViewModel

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewController.swift
mohamede1945: sounds good

Features/BookmarksFeature/Sources/CollectionsViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+                .sink { [weak self] in self?.updateMenuButton() }
+                .store(in: &cancellables)
+        }
+
+        private func updateMenuButton() {
+            if viewModel.editMode.isEditing {
+                navigationItem.rightBarButtonItem = UIBarButtonItem(
+                    title: l("button.done"),
+                    primaryAction: UIAction { [weak self] _ in
+                        self?.toggleEditing()
```
mohamede1945: let's not use toggle when the operation is not toggle, this operation should stop the editing and not depend on the existing value of editing. Same below for the edit button

Features/BookmarksFeature/Sources/CollectionsViewController.swift
```diff
@@ -0,0 +1,108 @@
...
+                    menu: UIMenu(children: menuActions)
+                )
+            }
+        }
+
+        private func toggleEditing() {
+            withAnimation {
+                viewModel.editMode = viewModel.editMode.isEditing ? .inactive : .active
+            }
+            updateMenuButton()
```
mohamede1945: I don't think we need this given the handling of objectWillChange above

mohamede1945 (changes requested):

Features/BookmarksFeature/Sources/BookmarksViewController.swift
```diff
@@ -6,20 +6,76 @@
...
 
 import Combine
+import FeaturesSupport
 import Localization
 import SwiftUI
 import UIx
 
+#if QURAN_SYNC
+    @MainActor
+    private final class BookmarksCollectionsNavigator {
```
mohamede1945: I don't understand why do we need this, could you please explain?

Features/BookmarksFeature/Sources/CollectionsView.swift
```diff
@@ -0,0 +1,133 @@
...
+
+    import Localization
+    import MobileSync
+    import NoorUI
+    import QuranKit
+    import SwiftUI
+    import UIx
+
+    @MainActor
+    struct CollectionsView: View {
```
mohamede1945: Could please rename CollectionsView, CollectionsViewModel, CollectionsViewController, CollectionsBuilder, etc to be AyahBookmarkCollections or BookmarkCollections? Since CollectionsView could be confused with the native CollectionView component

AhmedNMahran: It was originally added as a small bridge so SwiftUI actions could trigger UIKit navigation through a weak VC reference.

I removed it, because it added unnecessary indirection for a simple flow.

Now the builder injects the navigation closures directly into `BookmarksViewModel`, to keep same app behavior.

AhmedNMahran:
@mohamede1945 recheck please جزاك الله خيرا

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+    import UIKit
+
+    final class AyahBookmarkCollectionsViewController: UIHostingController<AyahBookmarkCollectionsView> {
+        // MARK: Lifecycle
+
+        init(
+            viewModel: AyahBookmarkCollectionsViewModel,
+            title: String = l("bookmarks.collections"),
+            allowsCollectionManagement: Bool = true,
+            allowsBookmarkDeletion: Bool = true
```
mohamede1945, line 21: Will these ever be false? If no, let's remove them please.

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -0,0 +1,108 @@
...
+    import SwiftUI
+    import VLogging
+
+    @MainActor
+    final class AyahBookmarkCollectionsViewModel: ObservableObject {
+        // MARK: Lifecycle
+
+        init(
+            ayahBookmarkCollectionService: AyahBookmarkCollectionService,
+            includedCollectionNames: Set<String>? = nil,
```
mohamede1945, line 19: what is includedCollectionNames used for? I understand excludedCollectionNames will be used for excluding old page bookmarks, colored collections, but do we really need includedCollectionNames?

Features/BookmarksFeature/Sources/BookmarksView.swift
```diff
@@ -15,8 +15,22 @@ import UIx
 @MainActor
 struct BookmarksView: View {
     @StateObject var viewModel: BookmarksViewModel
+    let showCollectionsAction: AsyncAction?
+
+    init(viewModel: BookmarksViewModel, showCollectionsAction: AsyncAction? = nil) {
```
mohamede1945: `showCollectionsAction: AsyncAction?` is already handled by the view model, so we don't have to pass it here too. Also this should simplify the code below, we just simply pass `showCollectionsAction: viewModel.showCollectionsAction`

mohamede1945: Also please remove canShowCollections

mohamede1945 (approved):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewController.swift
AhmedNMahran: done

PR #822, opened 2026-05-06
page bookmarks collections

AhmedNMahran:
Adds a standalone `Old Page Bookmarks` entry on the Bookmarks screen and syncs legacy page bookmarks into a fixed Mobile Sync collection with the same name.

- keeps the system collection logic out of `CollectionsViewModel`
- shows `Old Page Bookmarks` as its own row with the outlined bookmark icon
- excludes that collection from the general collections list
- adds unit coverage for creating the system collection and skipping already-synced ayahs

Screenshot page bookmark


Features/BookmarksFeature/Sources/OldPageBookmarksCollectionSync.swift
```diff
@@ -0,0 +1,114 @@
...
+    //
+
+    import AnnotationsService
+    import Foundation
+    import MobileSync
+    import QuranAnnotations
+    import QuranKit
+
+    @MainActor
+    struct OldPageBookmarksCollectionSync {
```
mohamede1945: As we discussed previously, we shouldn't do migration now

Core/Localization/Resources/en.lproj/Localizable.strings
mohamede1945, line 1: Arabic localization too please

mohamede1945 (changes requested):

Features/BookmarksFeature/Sources/OldPageBookmarksCollectionSync.swift
AhmedNMahran: I'm just taking getting the data and putting on the pages collection, not a real or complete migration yet.
can we leave this or otherwise can u suggest a specific change?

Core/Localization/Resources/en.lproj/Localizable.strings
AhmedNMahran: Did we agree to localize or just leave it?

mohamede1945: I don't know. But you added it to the localization files, so you already made the decision, right :) ?

Features/BookmarksFeature/Sources/OldPageBookmarksCollectionSync.swift
mohamede1945: what do you mean not a real or complete migration? could you explain please?

AhmedNMahran: I didn’t intend a destructive/real DB migration. The idea was:
- read legacy page bookmarks from the old local DB
- copy them into a fixed Mobile Sync collection named “Old Page Bookmarks”
- leave the old DB untouched
- skip already-copied ayahs

I can still remove this logic, but do u think we can still keep the ui logic, or the pr won't make sense and we close/draft it for now?

Core/Localization/Resources/en.lproj/Localizable.strings
AhmedNMahran: yes, I added to localiztion in en only so i don't hardcode it, but we need to know whether we will keep english only and user can edit or what should be done

Features/BookmarksFeature/Sources/OldPageBookmarksCollectionSync.swift
mohamede1945: The problem with this approach is:

1. Migrations should happen as part of onboarding or the upgrade flow, not inside individual screens.
2. The migration logic will likely run at specific points in time. Potentially once after the user upgrades rather than every time the user opens the bookmarks screen.

I think we can keep the PR, but remove the migration-related work from it. We should assume the migration has already happened, and this PR should focus only on displaying the data and supporting normal CRUD operations

AhmedNMahran: Agreed. will do insha'Allah,
جزاكم الله خيرا

AhmedNMahran:
@mohamede1945 recheck please :)

Features/BookmarksFeature/Sources/BookmarksView.swift
```diff
@@ -16,19 +16,29 @@ import UIx
 struct BookmarksView: View {
     @StateObject var viewModel: BookmarksViewModel
     let showCollectionsAction: AsyncAction?
+    let showOldPageBookmarksAction: AsyncAction?
 
-    init(viewModel: BookmarksViewModel, showCollectionsAction: AsyncAction? = nil) {
+    init(
+        viewModel: BookmarksViewModel,
+        showCollectionsAction: AsyncAction? = nil,
+        showOldPageBookmarksAction: AsyncAction? = nil
```
mohamede1945: Same as in https://github.com/quran/quran-ios/pull/821 the view model should be handling the action and I don't think we need the canShowOldPageBookmarks

mohamede1945 (approved):

Features/BookmarksFeature/Sources/BookmarksView.swift
AhmedNMahran: done, and I re-added `includedCollectionNames`, `allowsCollectionManagement`, and `allowsBookmarkDeletion` because #822 now uses them for the Old Page Bookmarks screen. They were unused in #821, but here they scope the screen to that collection and restrict its actions.

PR #823, opened 2026-05-06
add AyahBookmarkCollectionService.swift

AhmedNMahran:


Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,51 @@
...
+    import MobileSync
+    import QuranKit
+    import ReadingService
+
+    struct AyahBookmarkCollectionService {
+        // MARK: Lifecycle
+
+        init(quran: Quran = ReadingPreferences.shared.reading.quran) {
+            self.quran = quran
+        }
```
mohamede1945: Let's store the ReadingPreference instead of the Quran object otherwise users who change the mushaf will get the old value

mohamede1945 (approved):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,51 @@
...
+            AyahNumber(quran: quran, sura: Int(bookmark.sura), ayah: Int(bookmark.ayah))
+        }
+
+        func page(for bookmark: CollectionAyahBookmark) -> Page? {
+            ayah(for: bookmark)?.page
+        }
+
+        // MARK: Private
+
+        private let quran: Quran
```
mohamede1945: Actually this is not what I want it to be. I want the async sequence API to produce a new data type that has AyahNumber so consumers don't have to work with raw integer values of  sura and ayah

mohamede1945 (changes requested):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,91 @@
...
+                ayah: Int(bookmark.ayah)
+            ) else {
+                return nil
+            }
+            return AyahCollectionBookmark(bookmark: bookmark, ayah: ayah)
+        }
+    }
+
+    extension SyncService {
+        func removeBookmarkFromCollection(_ bookmark: AyahCollectionBookmark) async throws {
```
mohamede1945: you had a TODO here, could you please bring it back?

mohamede1945 (approved):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,91 @@
...
+    struct AyahBookmarkCollectionService {
+        // MARK: Lifecycle
+
+        init(readingPreferences: ReadingPreferences = .shared) {
+            self.readingPreferences = readingPreferences
+        }
+
+        // MARK: Internal
+
+        func collectionsSequence<S: AsyncSequence>(_ sequence: S) -> AsyncThrowingStream<[AyahBookmarkCollection], Error>
```
mohamede1945: This service is a wrapper of SyncService it is not a utility service. Consumers shouldn't need to depend on SyncService at all an this method should automatically call SyncService.collectionsWithBookmarksSequence to retrieve the data and then do the mapping. The way this is implemented makes AyahBookmarkCollectionService a utility rather than a data service.

After reviewing the other PR https://github.com/quran/quran-ios/pull/821 I found this problem, didn't see it earlier.

Also let's expose the other methods here, so consumers don't have to depend on SyncService at all

mohamede1945 (changes requested):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,121 @@
...
+            ) else {
+                return nil
+            }
+            return AyahCollectionBookmark(bookmark: bookmark, ayah: ayah)
+        }
+    }
+
+    private enum AyahBookmarkCollectionServiceError: Error {
+        case missingSyncService
+    }
```
mohamede1945: Ahmed, this service shouldn't work without a SyncService, what is the point of it when there is no sync service configured?

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,121 @@
...
+                        for try await collections in syncService.collectionsWithBookmarksSequence() {
+                            continuation.yield(self.collections(from: collections))
+                        }
+                        continuation.finish()
+                    } catch {
+                        continuation.finish(throwing: error)
+                    }
+                }
+                continuation.onTermination = { _ in task.cancel() }
+            }
```
mohamede1945: Use `.map` please

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,121 @@
...
+                bookmark: AyahBookmark(
+                    sura: bookmark.bookmark.sura,
+                    ayah: bookmark.bookmark.ayah,
+                    lastUpdated: bookmark.bookmark.lastUpdated,
+                    localId: bookmark.bookmark.bookmarkLocalId
+                )
+            )
+        }
+
+        func collections(from collections: [CollectionWithAyahBookmarks]) -> [AyahBookmarkCollection] {
```
mohamede1945: should be private

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,121 @@
...
+                AyahBookmarkCollection(
+                    collection: collection.collection,
+                    bookmarks: collection.bookmarks.compactMap(bookmark)
+                )
+            }
+        }
+
+        func page(for bookmark: AyahCollectionBookmark) -> Page {
+            bookmark.ayah.page
+        }
```
mohamede1945: Unused, please remove

AhmedNMahran:
@mohamede1945 recheck please, جزاك الله خيرا

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,114 @@
...
+
+        public func collectionsSnapshot() async throws -> [AyahBookmarkCollection] {
+            let sequence = syncService.collectionsWithBookmarksSequence()
+                .map { collections in
+                    self.collections(from: collections)
+                }
+            var iterator = sequence.makeAsyncIterator()
+            return try await iterator.next() ?? []
+        }
```
mohamede1945: Could you explain why we are moving away from async sequence you already had in the previous revision of the PR?

I see https://github.com/quran/quran-ios/pull/821/ already moved to use this pattern

AhmedNMahran: While testing, I found a runtime warning when collection updates arrived:

```swift
Publishing changes from background threads is not allowed
```

That happened because the async sequence can emit from a background thread, then downstream UI state was updated from that emission.

I first tried to fix by changing the service into a callback-style observer that delivered on `@MainActor`, but I see your concern that moved away from the async-sequence shape.

I’ll work on reverting to the previous async-sequence API shape and keep the fix insha' Allah:

AhmedNMahran: @mohamede1945 it should be fixed now insha'Allah please recheck:

One extra implementation detail: because the package targets iOS 15, I couldn’t expose the mapped sequence using typed `some AsyncSequence<[AyahBookmarkCollection], Error>`. I added a tiny type-erased `AyahBookmarkCollectionsSequence` wrapper so the service can still expose a typed async sequence without leaking the KMP sequence type or switching back to callbacks.

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        public func removeBookmarkFromCollection(_ bookmark: AyahCollectionBookmark) async throws {
+            try await syncService.removeAyahBookmarkFromCollection(bookmark.bookmark)
+        }
+
+        public func collectionsSequence() -> AyahBookmarkCollectionsSequence {
+            let sequence = syncService.collectionsWithBookmarksSequence()
+                .map { collections in
+                    self.collections(from: collections)
+                }
```
mohamede1945, line 94: This map retains self and could potentioally result in memory leak, use the static func collections instead and remove this method:

private func collections(from collections: [CollectionWithAyahBookmarks]) -> [AyahBookmarkCollection] {

mohamede1945 (approved):

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
AhmedNMahran: Done

PR #824, opened 2026-05-06
highlights feature

AhmedNMahran:
Implements sync-backed highlights using fixed Mobile Sync collections.

- treats `red`, `green`, `blue`, `yellow`, and `purple` collection names as read-only system highlight collections
- removes the dedicated highlight service and uses small helpers plus direct sync-gated wiring instead
- updates Quran highlight rendering and ayah menu highlight actions to read/write those collections
- shows highlight collections separately from editable custom collections
- adds localized highlight color names and helper test coverage
Screenshot 2026-06-08 at 4 00 26 PM


Core/Localization/Resources/ar.lproj/Localizable.strings
```diff
@@ -123,6 +123,17 @@
 "bookmarks.sync.body" = "سجّل الدخول إلى Quran.com للاحتفاظ بعلاماتك المرجعية متاحة عبر أجهزتك.";
 "bookmarks.sync.action" = "تسجيل الدخول";
 
+"bookmarks.collections" = "المجموعات";
```
mohamede1945: This is duplicated 3 times

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -65,7 +72,17 @@ final class AyahMenuViewModel {
         deps.quranContentStatePreferences.quranMode == .translation
     }
 
-    var highlightingColor: Note.Color { deps.noteService.color(from: deps.notes) }
+    var highlightingColor: HighlightColor {
+        #if QURAN_SYNC
+            if deps.ayahBookmarkCollectionService != nil {
+                return deps.highlightColor ?? .yellow
+            }
+        #endif
```
mohamede1945, line 84: Doesn't seem we actually need this code, the below code should be enough

Model/QuranAnnotations/HighlightColor.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  HighlightColor.swift
+//
+//  Created by Ahmed Nabil on 2026-05-06.
+//
+
+public enum HighlightColor: String, CaseIterable, Equatable, Hashable {
```
mohamede1945: Given we add this, we should remove Note.Color. Also don't make it String raw value representable, collection names are unlikely to be just "red", "green", etc

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -221,4 +267,116 @@ final class AyahMenuViewModel {
...
+        case .red: return .red
+        case .green: return .green
+        case .blue: return .blue
+        case .yellow: return .yellow
+        case .purple: return .purple
+        }
+    }
+
+    #if QURAN_SYNC
+        private func updateSyncedHighlight(color: HighlightColor) async throws {
```
mohamede1945, line 273: Similar to my comment in https://github.com/quran/quran-ios/pull/831#discussion_r3255610606, we shouldn't be reading from the database before updating the database in the feature code

Features/BookmarksFeature/Sources/BookmarksBuilder.swift
```diff
@@ -28,11 +30,16 @@ public struct BookmarksBuilder {
             let showOldPageBookmarksAction: (@MainActor (UIViewController) async -> Void)?
             if let syncService = container.syncService {
                 let ayahBookmarkCollectionService = AyahBookmarkCollectionService(syncService: syncService)
+                let prepareHighlightCollections: ([AyahBookmarkCollection]) async throws -> Void = { collections in
+                    try await HighlightBookmarkCollections.ensure(in: collections, using: ayahBookmarkCollectionService)
+                }
```
mohamede1945: As I mentioned several times before, please let's not add business logic in builders

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -221,4 +267,116 @@ final class AyahMenuViewModel {
...
+
+    private func legacyColor(for color: HighlightColor) -> QuranAnnotations.Note.Color {
+        switch color {
+        case .red: return .red
+        case .green: return .green
+        case .blue: return .blue
+        case .yellow: return .yellow
+        case .purple: return .purple
+        }
+    }
```
mohamede1945: Let's remove these after we use just a single color enum

Features/QuranViewFeature/QuranInteractor.swift
```diff
@@ -443,6 +447,15 @@ final class QuranInteractor: WordPointerListener, ContentListener, NoteEditorLis
...
     }
 
+    private func highlightColor(for verses: [AyahNumber]) -> HighlightColor? {
+        let colors = verses.compactMap { deps.highlightsService.highlights.highlightVerses[$0] }
+        guard colors.count == verses.count else {
+            return nil
+        }
+        let uniqueColors = Set(colors)
+        return uniqueColors.count == 1 ? uniqueColors.first : nil
+    }
```
mohamede1945: Let's move this logic to AyahMenu

UI/NoorUI/Features/AyahMenu/AyahMenuView.swift
```diff
@@ -322,42 +322,28 @@ private struct MenuGroup<Content: View>: View {
...
 private struct NoteCircles: View {
-    let selectedColor: Note.Color?
-    let tapped: @Sendable (Note.Color) async -> Void
+    let selectedColor: HighlightColor?
+    let tapped: @Sendable (HighlightColor) async -> Void
 
     var body: some View {
         HStack {
-            ForEach(Note.Color.sortedColors, id: \.self) { color in
+            ForEach(HighlightColor.allCases, id: \.self) { color in
```
mohamede1945: use sortedColors

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -93,7 +110,16 @@
...
         private let excludedCollectionNames: Set<String>
+        private let prepareCollections: ([AyahBookmarkCollection]) async throws -> Void
         private let navigateToPage: (Page) -> Void
+        private var didPrepareCollections = false
+
+        private static func highlightSortIndex(_ collection: AyahBookmarkCollection) -> Int? {
+            guard let color = HighlightColor(collectionName: collection.collection.name) else {
+                return nil
+            }
+            return HighlightColor.allCases.firstIndex(of: color)
```
mohamede1945: Use sortedColors

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -33,13 +36,27 @@
         @Published var collections: [AyahBookmarkCollection] = []
         @Published var collapsedCollectionIDs: Set<String> = []
 
+        static func sorted(_ collections: [AyahBookmarkCollection]) -> [AyahBookmarkCollection] {
+            collections.sorted { lhs, rhs in
+                switch (highlightSortIndex(lhs), highlightSortIndex(rhs)) {
+                case let (lhsIndex?, rhsIndex?):
+                    return lhsIndex < rhsIndex
+                case (_?, nil):
```
mohamede1945: use .some

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -33,13 +36,27 @@
...
+                    return true
+                case (nil, _?):
```
mohamede1945: use .some

Features/BookmarksFeature/Sources/HighlightBookmarkCollections.swift
```diff
@@ -0,0 +1,20 @@
+#if QURAN_SYNC
+    //
+    //  HighlightBookmarkCollections.swift
+    //
+    //  Created by Ahmed Nabil on 2026-05-10.
+    //
+
+    import QuranAnnotations
+
+    enum HighlightBookmarkCollections {
```
mohamede1945: I don't think it's a good idea to create a single use helper

mohamede1945 (changes requested):
Please make the bookmark icons colored with the collection color to visually differentiate between them

Features/AyahMenuFeature/AyahMenuViewModel.swift
AhmedNMahran: Agreed, added a todo for now

Model/QuranAnnotations/HighlightColor.swift
AhmedNMahran: May be this should be done on a follow-up after we now exactly how should we handle highlight naming, and for Note.Color, it is used by the legacy notes, so we should keep wdyt?

mohamede1945: I fail to see how the collection names is a deciding factor in whether we should have 1 enum or 2.

We should use a single enum shared by both core data and the new sync impl. Having 2 nearly identical enums that differ only in associated values increases maintenance overhead, adds unnecessary complexity and hard to reason about.

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -138,6 +153,16 @@ final class AyahMenuViewModel {
...
+        #if QURAN_SYNC
+            if selectedHighlightColor != nil, !containsText(deps.notes), deps.ayahBookmarkCollectionService != nil {
+                do {
+                    try await removeSyncedHighlights()
+                } catch {
+                    crasher.recordError(error, reason: "Failed to remove synced highlights")
+                }
+                return
+            }
+        #endif
```
mohamede1945, line 161: 1. delete notes when quran sync is enabled should basically just delete the notes not the highlights
2. we need to add a delete highlight menu option but maybe you have it in a different PR? it's good to have it in a different pr actually because we probably need more than just delete highlights, we need add to collection, remove from collection, create a collection, etc

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -194,15 +219,42 @@ final class AyahMenuViewModel {
 
     private let deps: Deps
 
+    // TODO: remove these after we use just a single color enum
```
mohamede1945: this pr is using a single color enum now, please impl the todo

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -43,11 +46,15 @@ final class AyahMenuViewModel {
         let sourceView: UIView
         let pointInView: CGPoint
         let verses: [AyahNumber]
-        let notes: [Note]
+        let notes: [QuranAnnotations.Note]
         let noteService: NoteService
         let textRetriever: ShareableVerseTextRetriever
-        let usesSyncedNotes: Bool
-        let noteCount: Int
+        let highlightVerses: [AyahNumber: HighlightColor]
```
mohamede1945: this seems like it should go under #if QURAN_SYNC below

Model/QuranAnnotations/Note.swift
```diff
@@ -20,17 +20,9 @@ public struct Note: Equatable {
...
-        case green
-        case blue
-        case yellow
-        case purple
-    }
-
     public let verses: Set<AyahNumber>
     public let modifiedDate: Date
-    public let color: Color
+    public let color: HighlightColor
```
mohamede1945: I think Color here and in the init should be gated by if not quran_sync enabled

Model/QuranAnnotations/QuranHighlights.swift
```diff
@@ -17,6 +17,7 @@ public struct QuranHighlights: Equatable {
     public var readingVerses: [AyahNumber] = []
     public var shareVerses: [AyahNumber] = []
     public var searchVerses: [AyahNumber] = []
+    public var highlightVerses: [AyahNumber: HighlightColor] = [:]
```
mohamede1945: Same here, it should be gated by quran_sync

Model/QuranAnnotations/Note.swift
mohamede1945: This actually might have a cascade effect (e.g. NoteService.color) as more methods will also have to be gated not quran_sync and that's good so we can clearly ensure note.color is not used in quran sync related code.

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -65,7 +72,12 @@ final class AyahMenuViewModel {
         deps.quranContentStatePreferences.quranMode == .translation
     }
 
-    var highlightingColor: Note.Color { deps.noteService.color(from: deps.notes) }
+    var highlightingColor: HighlightColor {
+        if let highlightColor = selectedHighlightColor {
+            return highlightColor
+        }
+        return deps.noteService.color(from: deps.notes)
```
mohamede1945: this should basically be #if #else quran_sync

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -194,15 +219,42 @@ final class AyahMenuViewModel {
 
     private let deps: Deps
 
+    // TODO: remove these after we use just a single color enum
+
+    private var selectedHighlightColor: HighlightColor? {
```
mohamede1945: this logic seems to be quran_sync specific so let's put the flag here to help us ensure we don't accidentally use this property in legacy persistence

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -151,12 +176,12 @@ final class AyahMenuViewModel {
         #endif
         let notes = deps.notes
         let color = deps.noteService.color(from: notes)
-        if let note = await _updateHighlight(color: color) {
+        if let note = await updateLegacyHighlight(color: color) {
             listener?.editNote(note)
         }
```
mohamede1945: this code should be in the #else of QURAN_SYNC otherwise we are storing highlights for notes even when quran sync is enabled

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -194,15 +219,42 @@ final class AyahMenuViewModel {
...
+                    try await updateSyncedHighlight(color: color)
+                    return nil
+                } catch {
+                    crasher.recordError(error, reason: "Failed to update synced highlights")
+                    return nil
+                }
+            }
+        #endif
+
+        return await updateLegacyHighlight(color: color)
```
mohamede1945: this should be in the else of the quran_sync otherwise we would update highlights in legacy stack even when quran_sync is enabled

Features/AyahMenuFeature/AyahMenuViewModel.swift
mohamede1945: This wasn't fixed. The code just moved one step down

Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift
mohamede1945, line 1: This is the code where 2 db operations happen to update or remove a highlight

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -104,5 +130,16 @@
                 return !excludedCollectionNames.contains(name)
             }
         }
+
+        private func ensureHighlightCollectionsIfNeeded(_ collections: [AyahBookmarkCollection]) async throws {
+            guard ensuresHighlightCollections, !didPrepareCollections else {
```
mohamede1945: why do we have 2 props controlling this logic?

UI/NoorUI/Components/List/NoorListItem.swift
mohamede1945, line 1: I'm not sure what are the changes here and why do we need them

UI/NoorUI/Theme/QuranHighlights+Theme.swift
```diff
@@ -38,6 +38,10 @@ extension QuranHighlights {
         // Sort order: share, reading, search, .note
         var versesByHighlights: [AyahNumber: UIColor] = [:]
 
+        for (verse, color) in highlightVerses {
+            versesByHighlights[verse] = color.uiColor.withAlphaComponent(Self.opacity)
+        }
+
         for (verse, note) in noteVerses {
             versesByHighlights[verse] = note.color.uiColor.withAlphaComponent(Self.opacity)
         }
```
mohamede1945: let's have #if quran_sync #else ... here please

Features/AyahMenuFeature/AyahMenuViewModel.swift
AhmedNMahran: this is handled in the next pr

Model/QuranAnnotations/Note.swift
AhmedNMahran: note coloring when highlighted is not implemented yet, so now in sync feature, if an ayah is in highlights and has a note, the note menu will show it unhighlighted, and the highlight menu (in bookmarks page) will show it under the relevant highlight collection.

Features/AyahMenuFeature/AyahMenuViewModel.swift
AhmedNMahran: I think the previous requirement was to remove Note.Color and unify using highlight color in both legacy and sync

Features/AyahMenuFeature/AyahMenuViewModel.swift
AhmedNMahran: the todo is done, it's a stale comment, will remove

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
AhmedNMahran: `ensuresHighlightCollections` is VM-level configuration injected at init, the main collections screen always sets it to true, filtered screens like `Old Page Bookmarks` don't. `didPrepareCollections` is a run-once guard because `start()` loops over an async sequence and receives repeated emissions. So: one is config, one is runtime state.

Features/AyahMenuFeature/AyahMenuViewModel.swift
mohamede1945: Are you answering 2nd point or both? because I think we should still do 1st point in this pr.

Model/QuranAnnotations/Note.swift
mohamede1945: I'm unsure I follow your point here. Let me clarify my comment and will wait to understand your point. My point is that when QURAN_SYNC is enabled note and highlight are separate while when it's disabled they are tight together in one entity.

So, what I suggest we should change this model to become:

```swift
struct Note {
#if QURAN_SYNC
    public let note: String
#else
    public let color: HighlightColor
    public let note: String?
#endif
}

Notice the note text become non-optional when quran sync enabled and no highlight color while keeping the old model when quran sync is disabled.

File: Features/AyahMenuFeature/AyahMenuViewModel.swift (hunk 18 above)
Comment (mohamede1945) [reply Features/AyahMenuFeature/AyahMenuViewModel.swift]: I'm unsure I understand how this relates to my comment, sorry

File: Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift (hunk 22 above)
Comment (mohamede1945) [reply Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift]: I think we can safely remove the config one to reduce complexity and no harm in having this logic for filtered screens, in fact one can land directly to them through deeplink in the future inshaa'Allah.

File: Features/BookmarksFeature/Sources/AyahBookmarkCollectionService.swift [hunk 24]
```diff
@@ -83,6 +83,33 @@
...
+                }
+            }
+
+            for ayah in ayahs where !targetAyahs.contains(ayah) {
+                try await addAyahBookmarkToCollection(
+                    collectionLocalId: targetCollection.collection.localId,
+                    ayah: ayah
+                )
+            }
+        }
```
mohamede1945: I don't think this logic is correct. If a user selects 2 ayahs where one of them in the collection and another one is not and decides to add both to the collection, this method will remove the one already in the collection.

Features/BookmarksFeature/Sources/AyahBookmarkCollectionsViewModel.swift
```diff
@@ -104,5 +127,16 @@
...
+        private func ensureHighlightCollectionsIfNeeded(_ collections: [AyahBookmarkCollection]) async throws {
+            guard !didPrepareCollections else {
+                return
+            }
+            didPrepareCollections = true
+            let existingNames = Set(collections.map(\.collection.name))
+            for name in HighlightColor.sortedColors.map(\.collectionName) where !existingNames.contains(name) {
+                try await ayahBookmarkCollectionService.createCollection(named: name)
+            }
+        }
```
mohamede1945: This code should live in the service layer.

Features/NotesFeature/NotesView.swift
```diff
@@ -86,24 +86,28 @@ private struct NotesViewUI: View {
     }
```
mohamede1945: Seems like the whole file can just be in if !QURAN_SYNC

Features/NotesFeature/NotesViewModel.swift
```diff
@@ -43,32 +43,40 @@ final class NotesViewModel: ObservableObject {
     @Published var searchTerm: String = ""
```
mohamede1945: Seems like the whole file can just be in if !QURAN_SYNC

Features/QuranViewFeature/QuranSyncedHighlightsObserver.swift
```diff
@@ -0,0 +1,84 @@
...
+
+        private func ensureHighlightCollectionsIfNeeded(_ collections: [AyahBookmarkCollection]) async throws {
+            guard !didPrepareCollections else {
+                return
+            }
+            didPrepareCollections = true
+            let existingNames = Set(collections.map(\.collection.name))
+            for name in HighlightColor.sortedColors.map(\.collectionName) where !existingNames.contains(name) {
+                try await ayahBookmarkCollectionService.createCollection(named: name)
+            }
```
mohamede1945: Same code as I mentioned earlier that needs to move to the data service so that it doesn't have to be repeated whenever we need to access collections

Features/QuranViewFeature/QuranInteractor.swift
```diff
@@ -261,14 +287,25 @@ final class QuranInteractor: WordPointerListener, ContentListener, NoteEditorLis
...
-        let notes = notesInteractingVerses(verses)
-        let input = AyahMenuInput(
-            sourceView: sourceView,
-            pointInView: point,
-            verses: verses,
-            notes: notes,
-            noteCount: syncedNoteCount(interacting: verses)
-        )
+        #if QURAN_SYNC
+            let notes: [Note] = []
```
mohamede1945: Why do we create this empty notes array?

Features/AyahMenuFeature/AyahMenuViewModel.swift
```diff
@@ -65,7 +75,13 @@ final class AyahMenuViewModel {
         deps.quranContentStatePreferences.quranMode == .translation
     }
 
-    var highlightingColor: Note.Color { deps.noteService.color(from: deps.notes) }
+    var highlightingColor: HighlightColor {
+        #if QURAN_SYNC
+            return selectedHighlightColor ?? .red
```
mohamede1945: I think we should use the preference to read the last used highlighted color

mohamede1945 (commented):
Jazak allah khyrn for the change, Ahmed. I've added a couple more comments. I'll work on them myself, don't worry about them for now.

mohamede1945 (approved):

PR #831, opened 2026-05-16
Update sync reading session in place

AhmedNMahran:
## Summary
- Update the active Mobile Sync reading session by `localId` when swiping pages.
- Send explicit timestamps through the new Mobile Sync API.
- Avoid adding/reordering unrelated recent reading sessions.

## Validation
- `QURAN_SYNC=1 xcodebuild -project Example/QuranEngineApp.xcodeproj -scheme QuranEngineApp -destination 'generic/platform=iOS Simulator' build`
- Previously validated on simulator with Maestro: selected session changed to latest page while the other sessions stayed unchanged.

https://github.com/user-attachments/assets/2e80d2a0-d76b-4a59-9e3d-7a255a8f988c

### DIFF
--- a/Domain/AnnotationsService/Sources/MobileSyncLastPageService.swift
+++ b/Domain/AnnotationsService/Sources/MobileSyncLastPageService.swift
@@ -8,6 +8,7 @@
 
 #if QURAN_SYNC
     import Combine
+    import Foundation
     @preconcurrency import MobileSync
     import QuranAnnotations
     import QuranKit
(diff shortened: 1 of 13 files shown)


Domain/AnnotationsService/Sources/MobileSyncLastPageService.swift
```diff
@@ -68,17 +50,96 @@
...
+            let firstVerse = toPage.firstVerse
+            let updatedSession = try await syncService.updateReadingSession(
+                localId: session.localId,
+                sura: Int32(firstVerse.sura.suraNumber),
+                ayah: Int32(firstVerse.ayah),
+                timestamp: Date()
+            )
+            updateState.setActive(localId: updatedSession.localId, page: toPage)
+            return LastPage(page: toPage, createdOn: updatedSession.lastUpdated, modifiedOn: updatedSession.lastUpdated)
         }
```
mohamede1945: I don't think this would be the proper implementation for this.

1. ReadingSessionUpdateState could dead lock in async-await context
2. The fact that we have to read all reading sessions from the database first and then update also doesn't look good to me.

I think a better approach is to make `func lastPages(quran: Quran) -> AnyPublisher<[LastPage], Never>` return `[ReadingSession]` and then update method would take the ReadingSession and the page to update

mohamede1945 (changes requested):

Domain/AnnotationsService/Sources/MobileSyncLastPageService.swift
AhmedNMahran: check again please @mohamede1945

mohamede1945 (approved):

PR #835, opened 2026-05-16
Add Mobile Sync notes service

AhmedNMahran:
## Summary
- Adds a thin Mobile Sync-backed notes service in `AnnotationsService`.
- Maps Mobile Sync notes to typed `SyncedNote` values with continuous ayah ranges.
- Supports create/update/delete by concrete sync note id.

## Scope
- Data/service only.
- No notes UI wiring.
- No legacy note persistence changes.
- No migration logic.

## Validation
- `QURAN_SYNC=1 xcodebuild build` via stacked UI branch.
- Non-sync `xcodebuild build` via stacked UI branch.
- Targeted `MobileSyncNoteServiceTests` passed.

### DIFF
--- a/Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
+++ b/Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
@@ -17,12 +17,6 @@
         public let modifiedDate: Date
 
         public var id: String { localId }
-
-        public var firstVerse: AyahNumber { startAyah }
-
-        public var verses: [AyahNumber] {
-            startAyah.array(to: endAyah)
-        }
     }
 
     public struct SyncedNotesSequence: AsyncSequence {
@@ -56,7 +50,7 @@
         private let makeIterator: () -> AsyncIterator
     }
 
-    public struct MobileSyncNoteService: @unchecked Sendable {
+    public struct MobileSyncNoteService {
         // MARK: Lifecycle
 
         public init(syncService: SyncService) {
--- a/Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
+++ b/Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
@@ -1,0 +1,133 @@
+#if QURAN_SYNC
+    //
+    //  MobileSyncNoteService.swift
+    //
+    //  Created by Ahmed Nabil on 2026-05-16.
+    //
+
+    import Foundation
+    @preconcurrency import MobileSync
+    import QuranKit
+
+    public struct SyncedNote: Equatable, Identifiable, Sendable {
+        public let localId: String
+        public let body: String
+        public let startAyah: AyahNumber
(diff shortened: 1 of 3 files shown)


Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
```diff
@@ -0,0 +1,133 @@
...
+    public struct SyncedNote: Equatable, Identifiable, Sendable {
+        public let localId: String
+        public let body: String
+        public let startAyah: AyahNumber
+        public let endAyah: AyahNumber
+        public let modifiedDate: Date
+
+        public var id: String { localId }
+
+        public var firstVerse: AyahNumber { startAyah }
```
mohamede1945: doesn't seem we need this

Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
```diff
@@ -0,0 +1,133 @@
...
+
+        public var verses: [AyahNumber] {
+            startAyah.array(to: endAyah)
+        }
```
mohamede1945: Is there a use case for using the list of verses a note span?

Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
```diff
@@ -0,0 +1,133 @@
...
+
+        public func makeAsyncIterator() -> AsyncIterator {
+            makeIterator()
+        }
+
+        private let makeIterator: () -> AsyncIterator
+    }
+
+    public struct MobileSyncNoteService: @unchecked Sendable {
```
mohamede1945: why this is @unchecked Sendable ?

mohamede1945 (approved):

Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
AhmedNMahran: Agreed. This was added as a UI convenience for the first ayah in the note range, but it does not belong in the data service model. Removed it and updated UI callers to use `startAyah` directly.

Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
AhmedNMahran: Not in the data layer. It was added to make UI range rendering easier, but that derived range should stay at the UI/use-site. Removed `verses` from `SyncedNote`; callers now derive it with `startAyah.array(to: endAyah)` when needed.

Domain/AnnotationsService/Sources/MobileSyncNoteService.swift
AhmedNMahran: It was added defensively because the wrapper stores `SyncService`, but the service itself does not need to conform to `Sendable`. Removed it to avoid an unnecessary unchecked concurrency annotation.

PR #843, opened 2026-06-02
Feature/streaming audio

OKKHALIL3:
Closes https://github.com/quran/quran-ios/issues/838

Adds a Stream Audio toggle in Settings that lets users play Quran recitation without downloading audio files first. When enabled,
AVPlayer streams directly from the remote URLs — no storage used.

How it works

Gapped reciters — streams all audio files directly, no download needed
Gapless reciters — the timing database (~88 KB) is downloaded once on first play so the app knows where each ayah starts in the
audio file. After that, subsequent plays are instant with no delays. The large audio files themselves are always streamed
Default is off — existing download-first behavior is completely unchanged
Implementation notes

AVURLAsset already handles https:// URLs natively, so QueuePlayer/AVPlayer needed zero changes
The streaming flag flows from AudioPreferences → AudioBannerViewModel → QuranAudioPlayer → the request builders, where it
switches between localURL and remoteURL
Added databaseDownloaded() and downloadDatabase() to QuranAudioDownloader for the gapless DB-only fetch path

### DIFF
--- a/Core/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Resources/ar.lproj/Localizable.strings
@@ -102,6 +102,8 @@
 "audio.select-end-verse" = "حدد نهاية التشغيل الصوتي";
 
 "audio.download-play-amount" = "تحميل وتشغيل الصوت لنهاية";
+"audio.streaming.title" = "تشغيل الصوت عبر الإنترنت";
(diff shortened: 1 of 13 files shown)


mohamede1945 (approved):
Looks great mashaa'Allah. Jazak Allah Khyrn

mohamede1945:
I'll try to do a release early next week inshaa'Allah. I saw brother @OKKHALIL3 (jazah Allah khyrn) created 2 more PRs, I'll review them later today and then maybe spend a day or 2 testing the app overall functionality and then send it for review.

PR #845, opened 2026-06-04
Add configurable delay between repetitions

### LINKED ISSUE #839: Add Configurable Delay Between Repetitions
Add a configurable delay that is applied before restarting a repeated verse range.

When a repetition cycle completes, the player should wait for the configured delay before beginning the next repetition. Users should be able to choose from a set of predefined delay options (No delay, 1s, 2s, 3s, 4s, or custom).

- **1s should be the default.**

**Motivation:**

This feature is intended to support Quran memorization (Hifz). When a verse range is repeated continuously without a pause, users may unintentionally associate the final verse of the range with the first verse of the next repetition cycle. Introducing a delay creates a clearer separation between repetitions, helping users focus on the intended verse sequence and reducing incorrect memorization patterns.

**Example:**

For a repetition range of verses 1–5 with a 3-second delay:

1. Verses 1–5 are played.
2. Playback pauses for 3 seconds.
3. Verses 1–5 are played again.
4. The cycle continues until the configured repetition count is reached or playback is stopped.

**Expected Behavior:**

* Users can configure a delay between repetition cycles.
* The delay is applied after the last verse in the repeated range and before the first verse of the next repetition.
* The delay applies to both finite repetition counts and infinite looping.

OKKHALIL3:
Closes #839

Adds a delay option between repetition cycles in the Advanced Audio screen.

## Changes

- New `RepetitionDelay` enum (`none`, `1s`, `2s`, `3s`, `4s`) in `QueuePlayer`
- `AudioPlayer` pauses the underlying player during the delay to prevent audio bleed-through, then restarts from the beginning of the range
- Delay is persisted in `AudioPreferences` with a default of 1s
- New "Delay Between Repetitions" section added to the Advanced Audio screen using the existing pill style
- Threaded through `AdvancedAudioOptions`, `AdvancedAudioOptionsViewModel`, and `AudioBannerViewModel`

## Testing

1. Open Advanced Audio and set a verse range with 2+ repetitions
2. Leave delay at 1s, play, and confirm a 1-second pause between cycles
3. Switch to "No delay" and confirm cycles loop without a gap
4. Tap pause during the delay window and confirm playback stops cleanly

### DIFF
--- a/Core/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Resources/ar.lproj/Localizable.strings
@@ -105,8 +105,8 @@
 "audio.streaming.title" = "تشغيل الصوت عبر الإنترنت";
 "audio.streaming.description" = "يوفر مساحة التخزين ببث الصوت، لكنه يتطلب اتصالًا بالإنترنت.";
 "audio.download-play-amount.description" = "عند بدء تشغيل الصوت، سيتوقف مع نهاية";
-"audio.repetition-delay" = "التأخير بين التكرارات";
-"audio.repetition-delay.none" = "بدون تأخير";
+"audio.repetition-delay" = "إيقاف مؤقت بين التكرارات";
+"audio.repetition-delay.none" = "٠ث";
 "audio.repetition-delay.1s" = "١ث";
 "audio.repetition-delay.2s" = "٢ث";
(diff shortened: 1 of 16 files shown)


mohamede1945 (approved):
Jazak Allah Khyrn for the changes

PR #846, opened 2026-06-04
Add custom repetition count for verse and range playback

### LINKED ISSUE #844: Allow Custom Repetition Count for Verse and Verse Range Playback
Replace the current **"Loop"** option in the Advanced Audio screen with a **"Custom"** repetition option that allows users to specify an exact number of repetitions for individual verses and verse ranges.

When selecting the repetition setting, users should see:

* **Loop (Infinite)** — current behavior
* **Custom Count** — selectable values from **1 to 100**

This feature is intended to support users who are memorizing the Quran and need precise control over repetition counts. For example, a user may want a verse to be repeated **30 times** and then automatically advance to the next verse. With the current infinite loop functionality, users cannot easily track how many repetitions have occurred.

**Expected Behavior:**

* Users can select a repetition count between **1 and 100**.
* The selected verse or verse range repeats exactly the specified number of times.
* Once the repetition count is reached, playback automatically proceeds according to the existing playback flow (e.g., next verse or next configured range).
* Infinite repetition remains available through the **Loop** option.

**Motivation:**
Provide better support for Quran memorization (Hifz) workflows by enabling users to practice verses a predetermined number of times without manually tracking repetitions.

OKKHALIL3:
Closes #844

Adds a custom repetition count option (1-100) to the Advanced Audio screen alongside the existing preset pills.

## Changes

- `Runs` enum gains a `custom(Int)` case with `maxRuns` returning the wrapped value
- `RunsChoicesSection` now shows a stepper (1-100) below the existing 1x/2x/3x/4x/5x/Loop pills
- Localization key `audio.repetition-count` added

## Testing

1. Open Advanced Audio and use the stepper to set a count above 5, e.g. 10
2. Play and confirm the verse/range repeats exactly 10 times then stops
3. Tap the Loop pill and confirm infinite looping still works
4. Tap a preset pill (e.g. 3x) and confirm it repeats 3 times


mohamede1945:
Jazak Allah khyrn.

Screenshot 2026-06-08 at 12 42 49 PM



I don't think I would do this. I need to keep the existing design as it's. Just change loop to be custom and then when user chooses custom, we add a new row with a way to select a custom value probably with a picker. Then add loop as part of the picker probably at the first top and at the very bottom


Also, it would be great to have an image or a video showing the change please.

Thank you!

OKKHALIL3:
Asalamu alykum. Updated per the requests. Here is a screen shot: 
Screenshot 2026-06-10 at 3 36 12 PM

mohamede1945 (approved):

mohamede1945:
Jazak Allah khyrn! I've modified it to be a menu directly to make it easier to understand (ux perspective) as my previous suggestion wasn't great, sorry.

Screenshot 2026-06-15 at 11 08 18 PM

<br />

simulator_screenshot_7ECC4924-C35C-46BE-8F25-4A89D5D0181D

PR #847, opened 2026-06-05
Add configurable delay between verses (multiplier of recited verse) (#840)

### LINKED ISSUE #840: Add Configurable Relative Delay Between Verses During Recitation
Add a configurable delay between verses during audio playback. Rather than using a fixed duration, the delay should be calculated as a multiplier of the duration of the verse that was just recited.

Example delay options:

* No delay (**default**)
* 0.25×
* 0.5×
* 0.75×
* 1×
* 1.25×
* 1.5×

For example, if a verse takes 8 seconds to recite and the user selects a 0.5× delay, the player would wait 4 seconds before starting the next verse.

**Motivation:**

This feature is intended to support Quran memorization (Hifz) by giving users time to mentally repeat, process, and recall each verse before the next one begins. Because the delay scales with the length of the verse, it provides a more natural memorization experience than a fixed delay.

While this feature is particularly valuable when combined with verse repetition, it should also work independently when verses are played only once.

**Examples:**

**Without repetition**

* Verse 1 is recited (8 seconds)
* 4-second delay (0.5× setting)
* Verse 2 is recited
* 4-second delay
* Verse 3 is recited

**With repetition**

* Verse 1 is recited
* 4-second delay
* Verse 1 is recited again
* 4-second delay
* Verse 1 is recited a third time
* 4-second delay
* Playback advances to Verse 2

**Expected Behavior:**

* Users can select a delay multiplier from predefined options.
* The delay is calculated using the duration of the previously recited verse.
* The delay is applied between every verse playback, including repeated verse playbacks.
* A "No Delay" option remains available.
* The feature works for both single-play and repeated-play configurations.
* The delay applies consistently regardless of reciter selection.

abdorizak:
Closes #840

Adds a **"Delay Between Verses"** option to the Advanced Audio screen. Unlike a fixed gap, the delay scales with the duration of the verse that was just recited (a multiplier), giving a more natural pause for memorization (Hifz) — time to mentally repeat and recall before the next verse begins.

## Behavior
- Delay = (recited verse duration ÷ playback rate) × selected multiplier.
- Applied between **every** playback: next verse, **repeated** verse, and new range runs.
- **No delay** is the default and stays available.
- Works for both single-play and repeated-play configurations.
- Consistent across reciters — both **gapped** and **gapless**.

## Options
`None` (default) · `0.25×` · `0.5×` · `0.75×` · `1×` · `1.25×` · `1.5×` · `1.75×` · `2×` · `2.25×` · `2.5×` · `2.75×` · `3×`

## Changes
- New `VerseDelay` enum in `QueuePlayer` (raw-value backed; cases appended in ascending order so saved prefs stay valid).
- `AudioPlayer` pauses the player and waits the computed delay before each onward playback; the **no-delay path is byte-for-byte the original** seamless gapless behavior. Pause/stop cancel a pending delay, and the internal pause is hidden from the UI.
- Threaded through `AudioRequest`, both request builders, `QuranAudioPlayer`, `AdvancedAudioOptions` / ViewModel / View, and `AudioBannerViewModel`.
- Persisted in `AudioPreferences` (default `None`) and restored when reopening the sheet.
- New "Delay Between Verses" pill section on the Advanced Audio screen (matches the playback-speed pill style).
- Localized title / "No delay" / footer in **all 16 languages**; numeric multiplier labels are `NumberFormatter`-formatted per locale (e.g. Arabic-Indic digits).
- Unit tests for the enum and the view-model wiring.

## Testing
1. Open Advanced Audio → scroll to **Delay Between Verses** → pick e.g. `0.5×` → play; confirm a pause ≈ half the just-played verse's length between verses.
2. Set "Play each verse" to 3× and confirm the delay also applies between repeats.
3. Compare a short vs long verse — the gap scales with verse length.
4. Try a gapped and a gapless reciter — same behavior.
5. Set a value, relaunch, and confirm it's remembered.

## Related
Independent of, but overlaps in the same files with:
- #845 (delay between *repetition cycles*, fixed seconds) — distinct feature; named `VerseDelay` vs its `RepetitionDelay` so they coexist.
- #846 (custom repetition count) — minimal overlap (view + strings).

Whichever of these merges later will need a small conflict resolution in `AudioRequest`, `AudioPlayer`, `AudioPreferences`, the options stack, and the strings files.

### DIFF
--- a/Core/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Resources/ar.lproj/Localizable.strings
@@ -113,7 +113,7 @@
 "audio.repetition-delay.5s" = "٥ث";
 "audio.repetition-delay.10s" = "١٠ث";
 
-"audio.verse-delay" = "التأخير بين الآيات";
+"audio.verse-delay" = "التوقف بعد كل آية";
 "audio.verse-delay.description" = "اختر مدة التوقف بعد كل آية. عند ١×، يتوقف التشغيل تقريبًا بقدر مدة الآية، مما يمنحك وقتًا للتكرار أو الحفظ قبل المتابعة.";
 
 // MARK: - Bookmarks
--- a/Core/Localization/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Resources/ar.lproj/Localizable.strings
@@ -105,6 +105,13 @@
 "audio.streaming.title" = "تشغيل الصوت عبر الإنترنت";
 "audio.streaming.description" = "يوفر مساحة التخزين ببث الصوت، لكنه يتطلب اتصالًا بالإنترنت.";
 "audio.download-play-amount.description" = "عند بدء تشغيل الصوت، سيتوقف مع نهاية";
+"audio.repetition-delay" = "مدة الإيقاف بين التكرارات (بالثواني)";
+"audio.repetition-delay.none" = "٠ث";
+"audio.repetition-delay.1s" = "١ث";
+"audio.repetition-delay.2s" = "٢ث";
+"audio.repetition-delay.3s" = "٣ث";
+"audio.repetition-delay.5s" = "٥ث";
+"audio.repetition-delay.10s" = "١٠ث";
 
 "audio.verse-delay" = "التأخير بين الآيات";
 "audio.verse-delay.none" = "بدون تأخير";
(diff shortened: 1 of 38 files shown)


mohamede1945:
Could you please resolve the merge conflicts there are lots of those? Thank you

Also please add video or images showing the new functionality. 

Jazak Allah khyrn

abdorizak:
> Could you please resolve the merge conflicts there are lots of those? Thank you
> 
> Also please add video or images showing the new functionality.
> 
> Jazak Allah khyrn

Let me check Akhii.

abdorizak:
@mohamede1945 here video please take a look it will expires with in 2 days https://streamable.com/5b9k46

mohamede1945 (approved):

PR #850, opened 2026-06-14
Upgrade mobile-sync to 0.1.10

ahmedre:


ahmedre:
note we can reduce churn in this pr by doing something like:

```swift
public typealias AuthService = SyncAuthService
```

inside of mobile-sync-spm - let me know if you prefer this, we'd need a new release which is easy.

btw, the reason this changed is part of the refactoring - the Kotlin `AuthService` today is the actual raw authentication module, and shouldn't be exposed to Android nor iOS directly. `SyncAuthService` is the facade Android and iOS should use. it does extra critical things that `AuthService` itself won't - like making sure logout properly stops sync, clears local data, etc.

mohamede1945 (approved):

PR #851, opened 2026-06-24
Update Localizable.strings

pulatovpro:
word mushaf was mistakenly translated by chatgpt as мушаф, i replaced all these to мусхаф


mohamede1945 (approved):
Jazak Allah Khyrn

PR #920, opened 2026-07-26
Confirm deletion of non-empty collections

mohamede1945:
## Summary

- require destructive confirmation before deleting a collection that contains bookmarks
- pass the pending collection through item-driven alert state so confirmation deletes the exact requested collection
- apply the same confirmation from the collection list and collection detail menu
- keep empty collection deletion immediate
- add English and Arabic confirmation copy
- cover both paths with MobileSync-backed regression tests

## Why

Collection delete actions previously called persistence directly regardless of collection contents. A swipe or menu action could therefore remove a populated collection without a chance to cancel.

## Impact

Users now receive a clear Cancel/Delete alert when a collection contains bookmarks. Empty collections retain the existing one-step deletion behavior. The pending collection is the single source of truth for alert presentation and the destructive action.

## Validation

- `make format-lint`
- `make test-sync TARGET=BookmarksFeature` — 48 tests passed
- `make build-no-sync TARGET=BookmarksFeature`
- `git diff --check`

### DIFF
--- a/Core/Localization/Sources/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Sources/Resources/ar.lproj/Localizable.strings
@@ -131,6 +131,8 @@
 "bookmarks.collections.new" = "مجموعة جديدة...";
 "bookmarks.collections.new.placeholder" = "اسم المجموعة";
 "bookmarks.collections.rename" = "إعادة التسمية";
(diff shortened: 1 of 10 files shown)


mohamede1945:
### Delete confirmation

Non-empty collection deletion confirmation

PR #944, opened 2026-08-01
Render Dhivehi translations in Faruma with correct Arabic shaping

Alix35301:
## Problem

Two defects in Dhivehi translations, one cosmetic and one not.

**1. الله renders with two shaddas.**

The translation stores the name correctly as fully vocalised Arabic:

```
ا  U+0627 ALEF
ل  U+0644 LAM
ل  U+0644 LAM
◌ّ U+0651 SHADDA
◌َ U+064E FATHA
ه  U+0647 HEH
```

The cause is font fallback. Arabic quotations inside a Dhivehi translation fall through to the system Arabic font (Geeza Pro on iOS), which applies the الله `rlig` ligature unconditionally. That ligature glyph already carries a shadda, so the text's own explicit shadda is drawn on top of it — two shaddas on the name of God.

Verified by glyph-run inspection on iOS 26.5. Geeza Pro collapses those 6 codepoints to 3 glyphs; fonts that treat the ligature as contextual leave 5 and render correctly:

| Correct (one shadda) | Doubled |
|---|---|
| Kitab, Al Bayan, Baghdad, DecoType Naskh, Damascus, Nadeem | Geeza Pro, Arial, Times New Roman, Muna, KFGQPC Hafs |

**2. Thaana has no coverage in the system font.**

iOS ships exactly one Thaana font — `NotoSansThaana-Regular`, in `/System/Library/Fonts/UnicodeSupport/` (confirmed by scanning the `cmap` of all 264 fonts in the runtime). It was redesigned away from the Faruma style Dhivehi readers expect; see notofonts/thaana#4, where a Dhivehi speaker calls the current design "practically unusable" for formal writing.

## Fix

- Bundle **Faruma** and select it for `dv` via a new `FontName.dhivehi` case.
- Give that font a **cascade list** to `Kitab-Regular` for Arabic. Kitab makes the الله ligature contextual, so vocalised Arabic renders with one shadda. The two fonts do not overlap — Kitab has zero Thaana glyphs — so the cascade order is unambiguous, and Kitab covers all 45 Arabic codepoints used in the translation.
- **The translation text is not modified.** Normalising away the explicit shadda would work around the font bug at the cost of corrupting search, sharing, and copy.
- A cascade requires a `UIFont`, and **SwiftUI does not apply Dynamic Type to a `Font` built from one** — the text size silently freezes. `textFont(for:)` therefore takes the size the caller is about to set on the environment and pre-scales via `UIFontMetrics`. Measured before/after at three sizes:

  ```
  before:  xSmall=31.0  large=31.0  accessibility3=31.0   (frozen)
  after:   xSmall=27.0  large=31.0  accessibility3=68.0   (scales)
  ```

- Match on the **primary language subtag** rather than the raw string, so a region or casing change from the translations endpoint cannot silently drop back to the system font.

Both surfaces that render Dhivehi text — body chunks and the footnote sheet — read `textFont`, so both are covered. Translator names stay on the system font deliberately: they are Latin for both Dhivehi translations, and Faruma's Latin glyphs are Times-derived.

## Also worth knowing

`KFGQPCHAFSUthmanicScript-Bold` — the Quran-text font — collapses the same ligature and doubles the shadda too. Same defect, different code path, unrelated to Dhivehi. Not addressed here.

## Testing

- `make test-no-sync` — full package test plan, passes
- `make build-sync` — passes, so the change compiles with and without `QURAN_SYNC`
- `make format-lint` — clean

Cascade routing verified in-simulator by glyph run rather than by eye:

```
arabic runs: [("Kitab-Regular", 5)]
thaana runs: [("Faruma", 6)]
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mohamede1945:
Salam alikom,

Jazak allah khyrn for the PR. But I don't understand the problem and without screenshots of what's wrong and what is the correct impl. It's really hard for me to see the problem.

Also would really appreciate if you can let me know the source of the font being bundled here.

Alix35301:
Assalam Alaikum

Faruma is an official font that was created by President office of the Maldives to replace the old Dhivehi font that used in Dhivehi texts. It is widely available in internet to for anyone to download and use. Almost all mobile apps, web apps in Maldives use this font to render Dhivehi text instead of system fallback font.

The font bundled was sourced from https://dhivehi.mv/fonts/ 

The current font that is being used is not very readable so the change forces app to use Faruma font for Dhivehi texts

For comparison

Faruma.ttf
image


IOS fallback font
image


Jazākallāhu Khayran

mohamede1945:
Wa alaikum assalam,

JazakAllahu khayran for the explanation, screenshots, and effort on this PR.

After examining the screenshots closely, I'm unable to verify the reported double-shadda defect. The iOS fallback screenshot shows one shadda above it, not two shaddas. Additionally, the second occurrence of الله in the Faruma screenshot—within وَاللهُ أَعْلَمُ—appears visually similar to the fallback rendering. Therefore, the screenshots do not demonstrate that the proposed font cascade fixes the reported issue consistently.

The main visible difference is the style of the Thaana typeface. I understand that Faruma may be preferred by Dhivehi readers, but the evidence provided does not establish that the existing iOS rendering is technically incorrect.

There is also a licensing blocker. The bundled Faruma file identifies itself as © The President’s Office 2002, “All rights reserved,” and states that its Latin glyphs are derived from Times New Roman. I could not find explicit permission allowing the font to be redistributed and embedded in the app. Public download availability alone does not grant those rights.

Given the unclear reproduction and the font-licensing concern, I’m going to close this PR. We can reconsider it if explicit redistribution rights and a clear reproduction of the rendering defect are provided.

JazakAllahu khayran again for your contribution.

image

PR #987, opened 2026-08-16
chore(ci): add macOS workflow guardrails

yzua:
## Why

Recent iOS CI bursts queued for hours because every `main` push kept a separate macOS run alive. One package job also remained active for roughly six hours because no timeout was configured. This keeps iOS on GitHub-hosted `macos-26`; it does not move any job to QF self-hosted runners.

## Changes

- cancel superseded runs for the same branch or pull request
- add 15–45 minute job timeouts based on normal observed runtimes
- set default workflow permissions to `contents: read`
- upgrade and SHA-pin `actions/checkout` and Codecov to Node-24-compatible releases

## Validation

- `git diff --check`
- `nix run nixpkgs#actionlint -- -color=false .github/workflows/ci.yml`

This is intentionally PR-only for iOS team review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,12 +6,20 @@ on:
       - main
   pull_request:
 
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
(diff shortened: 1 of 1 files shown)


mohamede1945 (approved):
Looks great, jazak Allah khyrn

PR #997, opened 2026-08-17
ci: bump actions/checkout to v4

sporteka2:
`actions/checkout@v2` runs on the deprecated Node 16 runtime; GitHub now warns on every run and will eventually brown it out. v4 is the current major (runs on Node 20). No other changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: macos-26
 
     steps:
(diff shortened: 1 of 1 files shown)


mohamede1945:
Thanks for the contribution! This was superseded by https://github.com/quran/quran-ios/pull/987, which upgraded `actions/checkout` to SHA-pinned v7.0.1 across the workflow. Closing since no further changes are needed.

PR #1005, opened 2026-08-22
fix: use IndoPak sura names

mohamede1945:
## Summary

- use reading-specific names for the five suras that differ in the IndoPak mushaf
- add fully marked Arabic names and accented English transliterations to `Localizable.strings`
- fall back to English for other localizations while keeping Madani names unchanged
- render every IndoPak Arabic sura name with the bundled Noorehira font for a consistent style

## Testing

- `make format-lint`
- `make test-no-sync TARGET=LocalizationTests`
- `make test-sync TARGET=LocalizationTests`
- `make test-no-sync TARGET=QuranLocalizationTests`
- `make test-sync TARGET=QuranLocalizationTests`
- `make test-no-sync TARGET=NoorUITests`
- `make test-sync TARGET=NoorUITests`
- `plutil -lint` for English and Arabic `Localizable.strings`
- `git diff --check`

### DIFF
--- a/Core/Localization/Sources/Resources/ar.lproj/Localizable.strings
+++ b/Core/Localization/Sources/Resources/ar.lproj/Localizable.strings
@@ -20,6 +20,13 @@
 "surah" = "السورة";
 "quran_alquran" = "القرآن";
 
+// IndoPak sura names
+"quran.sura-name.indopak.17" = "بَنِي إِسْرَائِيل";
+"quran.sura-name.indopak.40" = "الْمُؤْمِن";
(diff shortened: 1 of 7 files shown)


ahmedre:
do we want to also update the English translations for naskh also? the problem is not all of them are translated though

ahmedre:
update - on android decided to fall back to the originals for them for now because they weren't all translated and I don't know if we have a source to get the correct translations

mohamede1945:
I don't use any sura name translations yet. But something I'm planning to do soon might be today 😁

Issue #195, opened 2017-07-28
deep linking


sabeurthabti:
Salam, 

I am working the quranic audio iOS/Android app. I want to link to the iOS application (when possible), does this application implement deep linking?

thanks

mohamede1945:
Salam,

No, we don't support any deep/universal linking. But let us know what you would like us to support and we will be more than happy to implement them.

sabeurthabti:
salam, 

I want to able to link to `https://quran.com/1` or `quran://chapter/1` and when the app is available it will open to that given surah.

mohamede1945:
I think we can easily handle https://quran.com/1 and the web app will only need to upload a small JSON file `apple-app-site-association`, check the documentation for iOS here.

https://developer.apple.com/library/content/documentation/General/Conceptual/AppSearch/UniversalLinks.html

Also, I think we should handle going straight to an Ayah/Verse as in https://quran.com/113/2

sabeurthabti:
@mohamede1945 That would be amazing, I will raise a PR for those paths. Can you provide me with the appID, it should look something like this: `2TDUX39LX8.com.reddit.Reddit`. 

How to get the appID https://blog.branch.io/what-is-an-aasa-apple-app-site-association-file/

sabeurthabti:
@mohamede1945  any update regarding the appID?

sabeurthabti:
No worries @mohamede1945 - Thank you. 
I pushed the path and it can be confirmed here (http://staging.quran.com/apple-app-site-association). 

Would it possible to support quran.com and staging.quran.com ?

I am available to do any testing whenever you require it. 

thanks.

sabeurthabti:
I would like to push for this again.  For us to test it we need to support `staging.quran.com` - I will try to help get this done in the coming days. I think the endpoint `/apple-app-site-association` has problems, which I will fix.

mohamede1945:
Thanks a lot and sorry for late reply. I went through a lot in the last 2 weeks.

sabeurthabti:
no worries bro - inshallah you're well :( 🙌

mohamede1945:
Thank you :)

PR #1030, opened 2026-09-06
feat: navigate to sura and ayah deep links

### LINKED ISSUE #348: Allow other apps to launch this app and jump to a particular Ayah

A-Levin:
Closes #348. Part of #195.

### What

`quran://` and `quran-ios://` were registered in #657, but `LaunchBuilder.navigateTo(path:)` was left as `// TODO: Implement the actual navigation`, so a link opened the app and stopped on whatever screen was last shown. This implements the navigation.

Supported links follow the format `QuranForwarderActivity` already handles in Quran for Android:

- `quran://2` opens sura 2
- `quran://2/255` opens ayah 2:255 and highlights it
- `quran-ios://` keeps working as an alias

### Parsing

`QuranDeepLink` maps a `URL` to either a sura or an ayah:

- the sura is read from the URL host, since `quran://2/255` parses as host `2` and path `/255`. The previous stub read `url.path` only, so the sura number was dropped
- non numeric segments are skipped, matching Android, so `quran://sura/2/255` resolves like `quran://2/255`
- the scheme is matched case insensitively
- validation goes through `Sura` and `AyahNumber` against the reading currently selected in `ReadingPreferences`. Anything out of range returns `nil` and the link is ignored rather than clamped to a nearby verse

Unit tests cover valid links, both schemes, trailing slashes, extra segments, `1:1` and `114:6`, the last ayah of a sura, and the rejected cases: sura `0`, `115`, `999`, ayah `0`, `2:287`, `1:8`, foreign schemes, and links with no sura at all.

### Navigation

- the entry point moves from `LaunchBuilder` to `LaunchStartup`. `LaunchBuilder` is a stateless struct, while `LaunchStartup` owns the built view controller hierarchy, which is what navigation needs. This is a source breaking change for anyone calling `LaunchBuilder.handleIncomingUrl(urlContext:)`; happy to keep a deprecated forwarding method instead if you prefer
- a link that arrives before the UI exists, for example during a migration or while waiting for protected data, is stored and replayed once `showApp` finishes
- `AppViewController.navigate(to:)` dismisses any presented modal, selects the home tab, pops it to its root, then pushes the Quran view. Popping first keeps repeated links from stacking Quran screens
- `Example` now also reads `connectionOptions.urlContexts` in `scene(_:willConnectTo:options:)`. A link that launches the app is delivered there and not to `scene(_:openURLContexts:)`, which is likely why #348 reports that nothing happens when the app is not already running

### Notes

- audio is out of scope here. As a follow up I would like to carry an optional playback range and repeat counts in the link, for example `quran://2/255?to=2:257&play=true&verse_repeat=3&range_repeat=infinite`, so a memoriser can open the app already looping a range. The engine already has `AdvancedAudioOptions` and the `audioBanner?.play(from:to:repeatVerses:)` seam, so it would mostly be threading an optional request through `QuranInput`. Let me know if you would take that
- universal links are not addressed. `https://quran.com/apple-app-site-association` currently returns 404, so `quran.com` links cannot be claimed yet
- the shipped app is not in this repository. If its target does not register these schemes and forward to `handleIncomingUrl`, this will work in `QuranEngineApp` but not in the App Store build

### Testing

`make test-no-sync`, `make test-sync`, `make build-example-no-sync`, `make build-example-sync` and SwiftFormat all pass on CI.

Verified the new tests actually fail when the logic breaks, by running CI against three deliberately broken copies of the parser: one with the scheme check removed, one with the ayah range check removed, one reading the path without the host.

### DIFF
--- a/Example/QuranEngineApp/Classes/SceneDelegate.swift
+++ b/Example/QuranEngineApp/Classes/SceneDelegate.swift
@@ -27,6 +27,11 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         launchStartup?.launch(from: window)
 
         self.launchStartup = launchStartup
+
+        // A link that launches the app arrives here, not in `scene(_:openURLContexts:)`.
+        if let urlContext = connectionOptions.urlContexts.first {
+            launchStartup?.handleIncomingUrl(urlContext: urlContext)
+        }
     }
 
     func sceneDidDisconnect(_ scene: UIScene) {
@@ -61,7 +66,7 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
         guard let urlContext: UIOpenURLContext = URLContexts.first else {
             return
         }
-        launchBuilder?.handleIncomingUrl(urlContext: urlContext)
+        launchStartup?.handleIncomingUrl(urlContext: urlContext)
     }
 
     // MARK: Private
--- a/Features/AppStructureFeature/Sources/App/AppBuilder.swift
+++ b/Features/AppStructureFeature/Sources/App/AppBuilder.swift
@@ -13,7 +13,7 @@ import UIKit
 struct AppBuilder {
     let container: AppDependencies
 
-    func build() -> UIViewController {
+    func build() -> AppViewController {
         let interactor = AppInteractor(
             supportsCloudKit: container.supportsCloudKit,
             analytics: container.analytics,
@@ -26,10 +26,9 @@ struct AppBuilder {
                 SettingsTabBuilder(container: container),
             ]
         )
-        let viewController = AppViewController(
+        return AppViewController(
             analytics: container.analytics,
             interactor: interactor
         )
-        return viewController
     }
 }
--- a/Features/AppStructureFeature/Sources/App/AppViewController.swift
+++ b/Features/AppStructureFeature/Sources/App/AppViewController.swift
@@ -20,6 +20,8 @@
 
 import Analytics
 import Crashing
+import FeaturesSupport
+import QuranKit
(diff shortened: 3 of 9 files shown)


mohamede1945:
The code is very well written, jazak Allah khyrn. I'll merge it after the CI passes inshaa'Allah.

mohamede1945 (approved):

PR #1034, opened 2026-09-07
fix: exclude downloaded resources from device backups

### LINKED ISSUE #675: Exclude downloaded audio from iCloud Backup
When iCloud Backup is enabled on a device, the backup it creates will include recitation/translation audio downloaded by the application, causing the device’s backup size to be significantly higher than it would have been without the audio data. Currently, the only way to mitigate this is to turn off backup of the app’s data entirely, in iCloud Backup’s Details screen, as pictured below.

Since this data is reproducible (i.e. by tapping the “download” button), this does not need to be part of a device’s backup. While we can simply apply the workaround described earlier, which is to turn off backup of the app’s data, ideally the app itself should prevent audio from being part of backups. Users that want to back their device up to iCloud but are mindful of iCloud storage consumption should not have to manually exclude the app from their device’s backup. Preventing audio from being backed up will also greatly reduce redundant network traffic as significantly less data will be uploaded by the system by default. (It is also worth noting that iCloud Backup is enabled by default on devices signed into iCloud.) Also, doing this will reduce the time it takes for a new device to finish restoring from iCloud on slower internet connections.

Foundation provides a way to mark certain files or directories created by our application as excludable from iCloud Backup, so that all data, except for that which is marked as excluded, will be included in device backups ([see article “Optimizing Your App’s Data for iCloud Backup” at developer.apple.com](https://developer.apple.com/documentation/foundation/optimizing_your_app_s_data_for_icloud_backup)). Quran recitation/translation audio most closely fits the article’s description of “non-purgeable” data. To prevent backing up non-purgeable data, it advises applying the `isExcludedFromBackup` resource value to the URL of a given file or directory. Assuming this is applicable to this application and the way it stores audio files, this may be a solution to the problem.

Screenshot of iCloud Backup details of an iPhone, highlighting the large size of the app’s backup data

A-Levin:
Fixes #675. Downloaded audio, translations, and reading resources can be downloaded again, but their Documents directories were included in device backups.

A migration registered for release **2.6.9** marks `audio_files`, `translations`, and `readings` as excluded from backup. It creates these roots if needed because older filesystem migrations can populate them concurrently. Existing downloaded contents are preserved, and a failure for one directory does not prevent processing the others.

For new downloads, the shared batch downloader marks each destination directory as excluded when preparing it to receive the downloaded file. This covers audio and timing databases, translations, reading resources, and directories recreated after deletion. Files extracted into those directories are covered by the directory exclusion. Attribute failures are reported without blocking the download.

The implementation uses the existing `FileSystem` boundary and removes the per-launch audio exclusion task.

Validation:
- All 12 `BatchDownloaderTests` passed with sync enabled and disabled.
- All 4 `AppMigrationFeatureTests` passed with sync enabled and disabled.
- Regression tests inspect real filesystem backup attributes, verify existing contents are preserved, and cover directory recreation, repeated migration, and continuation after a directory failure.
- `make format-lint` and `git diff --check` passed.

An actual iCloud backup/restore was not exercised locally.


mohamede1945:
Jazak Allah Khyrn for working on this. I updated the PR to use a migration for existing installs and apply backup exclusion when the shared downloader creates destination directories. This avoids repeating the work on every launch and covers directories recreated after deletion.

I also extended the exclusion to translations and reading resources, since they use the same download path and can be downloaded again.

mohamede1945 (approved):

A-Levin:
Wa iyyak

PR #1035, opened 2026-09-07
fix: normalize search input from Urdu, Farsi and Pashto keyboards

### LINKED ISSUE #841: Search not working
Searching for any Arabic word only works work with Arabic keyboard. It should work with keyboards that have same letters with Arabic like Urdu Farsi Pashto and etc.

A-Levin:
Fixes #841. Urdu, Farsi, and Pashto keyboards can produce letters such as `ک` and `ی` where Quran text uses `ك` and `ي`, causing submitted searches to miss matching verses.

The shared search similarity table now covers the relevant keheh, yeh, heh, and ta marbuta forms, including Urdu yeh barree with hamza (`ۓ`). These mappings are tolerant search substitutions, not linguistic equivalences. Each mapping retains the original character so it can still match translation text. Unrelated consonants such as `پ`, `چ`, `ژ`, and `گ` are not mapped to Arabic lookalikes. The SQL wildcard character class is derived from the mapping keys to keep candidate retrieval and regex matching consistent.

Submitted searches now apply canonical Unicode composition before removing marks. This makes composed and decomposed spellings of the same character behave consistently, while preserving the original query for display. Result matching also tries canonically composed text. Autocomplete retains its existing query normalization and database lookup behavior; extending its keyboard support is outside this change. Translation-search routing is unchanged.

Validation:
- All 39 `QuranTextKitTests` passed with sync disabled and enabled, run sequentially to avoid interference through the shared translation fixture directory.
- Regression tests cover equivalent Unicode encodings, accented Latin input, original translation letters, unrelated consonants, and end-to-end Quran search results.
- Existing autocomplete snapshots and the explicit autocomplete-normalization regression test passed.
- `make format-lint` and `git diff --check` passed.

### DIFF
--- a/Domain/QuranTextKit/Sources/Search/Searchers/CompositeSearcher.swift
+++ b/Domain/QuranTextKit/Sources/Search/Searchers/CompositeSearcher.swift
@@ -51,7 +51,7 @@ public struct CompositeSearcher {
     // MARK: Public
 
     public func autocomplete(term: String, quran: Quran) async -> [SearchText] {
-        guard let term = SearchTerm(term) else {
+        guard let term = SearchTerm(term, normalizeUnicode: false) else {
             return []
         }
         logger.info("Autocompleting term: \(term.compactQuery)")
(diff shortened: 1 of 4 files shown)


mohamede1945 (approved):

mohamede1945:
Amazing work, jazak Allah khyrn!

A-Levin:
Wa iyyak. Thank you, brother, I'm glad to help.

PR #1036, opened 2026-09-07
feat: remember verse and range repeat counts between playbacks

### LINKED ISSUE #678: Persist more audio player settings
Currently, AFAIK, only the choice of the reciter is persisted from this screen.

Can we also persist the choices in the screen for the options "Play each verse" and "Play set of verses"? Just saving them in user defaults could do the job for now, so that whenever user plays, it is the same choice from the previous time they played.

IMG_5954

Every single time, I find myself modifying it to play each verse 3 times and the entire range 3 times. Would be better if it is persisted.

A-Levin:
Fixes #678.

## Behavior

Remember verse and range repeat counts between playback sessions. Tapping Play in Advanced Audio Options saves both counts alongside the existing delay settings; Cancel discards edits. Normal playback reads the saved counts, and opening Advanced Audio Options after relaunch initializes its counts from those preferences. The long-press Repeat action still forces an endless range.

## Preference storage

Keep integer conversion in AudioPreferences: positive integers represent finite counts, zero represents endless repetition, and missing or negative stored values default to one run. TransformedPreference supports optional write validation, enabled for repeat counts, so invalid finite counts leave the existing preference unchanged. Existing transformers continue accepting writes by default.

## Validation

- All 38 QuranAudioKit tests passed locally in both sync and no-sync modes.
- Six preference tests exercise actual storage: defaults, finite and endless round trips, negative stored values, and invalid writes preserving existing or missing preferences.
- Formatting lint passed.
- AudioBannerViewModel startup wiring was reviewed but is not covered by a UI regression test.

An easy reset for switching from memorization to ordinary listening remains outside this change.


mohamede1945 (approved):
Looks great, jazak Allah khyrn

A-Levin:
Wa iyyak

PR #1038, opened 2026-09-07
fix: use available simulator runtimes for tests and example launches

### LINKED ISSUE #1037: CI: Package jobs fail intermittently because the pinned simulator runtime is missing on some runner images
Assalamu alaikum wa rahmatuLlahi wa barakatuh,

`Package` and `PackageSync` fail intermittently with an error that has nothing to do with the change under test:

```
xcodebuild: error: Unable to find a device matching the provided destination specifier:
```

## Why

`Makefile` pins both the simulator and its runtime:

```make
PACKAGE_DESTINATION ?= name=iPhone 17e,OS=26.5
```

The `macos-26` label is currently handing out more than one runner image version, and the older one does not have the iOS 26.5 runtime installed. Device types come from the Xcode the job selects, but runtimes come from the image, so a job that lands on an older image cannot satisfy `OS=26.5`.

This is visible in a single run — https://github.com/quran/quran-ios/actions/runs/34165095745, where four jobs got one image and `PackageSync` got an image from a month earlier:

| Job | Runner image version | Result |
|---|---|---|
| ExampleApp | `20260831.0337.3` | pass |
| ExampleAppSync | `20260831.0337.3` | pass |
| Package | `20260831.0337.3` | pass |
| SwiftFormat | `20260831.0337.3` | pass |
| PackageSync | `20260728.0273.1` | **fail** |

`Package` ran the very same destination string on the same commit and passed, so it is the image the job happened to draw, not the job or the code. The example app jobs are immune because `EXAMPLE_DESTINATION` is already `generic/platform=iOS`.

I have seen this three times in the last day, on https://github.com/quran/quran-ios/pull/1034 and https://github.com/quran/quran-ios/pull/1036.

## Suggested fix

Let the destination take whatever runtime the image has:

```make
PACKAGE_DESTINATION ?= name=iPhone 17e,OS=latest
```

The device model stays pinned, so tests still run on a known screen size and idiom; only the exact runtime becomes whatever is installed. PR to follow.

## Two things I could not investigate

**The error is undiagnosable from the logs.** `xcodebuild` prints the list of destinations it *did* find right after that error, but the output goes through `| xcbeautify`, which drops it. That is why the table above is built from runner image versions rather than from the actual list of available runtimes. A `xcrun simctl list devices available` step before the tests would make these failures readable. Happy to add it if you want it.

**A second, unrelated flake.** In https://github.com/quran/quran-ios/actions/runs/34163549520, `PackageSync` failed differently:

```
CoreData: warning: 'MO_PageBookmark' (0x1068b4900) from NSManagedObjectModel (0x106ab5450) claims 'MO_PageBookmark'
CoreData: warning: 'MO_PageBookmark' (0x1068b6f40) from NSManagedObjectModel (0x109b3a670) claims 'MO_PageBookmark'
error: +[MO_PageBookmark entity] Failed to find a unique match for an NSEntityDescription to a managed object subclass
```

`CoreDataPageBookmarkPersistenceTests.testRemovePageBookmarksDeletesEveryRequestedPage` and `CoreDataLastPagePersistenceTests.testUpdateCollapsesEverySourcePageIntoDestination` failed there. Every job in that run had the same image, so this one is not about runners: the same managed object model is being loaded twice in the test process. I have not tried to fix it, and I am mentioning it here only so the two are not confused with each other. Say the word if you would like me to look into it.

A-Levin:
Fixes #1037: package CI can fail before building when a runner image does not have the pinned iOS 26.5 simulator runtime.

Package builds and tests now use `name=iPhone 17e,OS=latest`. Example launch targets also default to the newest available installed iOS runtime, keeping the existing iPhone 17e and iPhone 17 Pro Max device choices. Since `simctl` requires a concrete version for device lookup, the example defaults resolve that version from installed runtimes and pass it to both the build destination and launcher. Explicit simulator and destination overrides remain supported.

Validation:
- Checked all five example launch command expansions, including local sync and What's New flows.
- Verified numeric version ordering (26.10 after 26.5), exclusion of unavailable and non-iOS runtimes, and explicit version overrides using a temporary `xcrun` fixture.
- Verified both default example launch commands resolve to iOS 26.5 on a machine with iOS 18.6, 26.0, and 26.5 installed.

The original failure depends on which runner image CI receives. A green run alone does not verify behavior on the older image; the evidence and affected runs are documented in #1037.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ BUILD_TOOLS_DIR=./BuildTools
 
 PACKAGE_SCHEME ?= QuranEngine-Package
 PACKAGE_SDK ?= iphonesimulator
-PACKAGE_DESTINATION ?= name=iPhone 17e,OS=26.5
+PACKAGE_DESTINATION ?= name=iPhone 17e,OS=latest
 EXAMPLE_PROJECT ?= Example/QuranEngineApp.xcodeproj
 EXAMPLE_SCHEME ?= QuranEngineApp
 EXAMPLE_SDK ?= iphonesimulator
--- a/Makefile
+++ b/Makefile
@@ -10,8 +10,10 @@ EXAMPLE_SCHEME ?= QuranEngineApp
 EXAMPLE_SDK ?= iphonesimulator
 EXAMPLE_DESTINATION ?= generic/platform=iOS
 EXAMPLE_BUNDLE_IDENTIFIER ?= com.quran.QuranEngineApp
-EXAMPLE_NO_SYNC_SIMULATOR ?= iPhone 17e,26.5
-EXAMPLE_SYNC_SIMULATOR ?= iPhone 17 Pro Max,26.5
+# simctl needs a concrete runtime version when locating the device to launch.
+LATEST_IOS_SIMULATOR_OS = $(shell xcrun simctl list runtimes | sed -nE 's/^iOS ([0-9.]+) .* - com\.apple\.CoreSimulator\.SimRuntime\.iOS-[0-9-]+$$/\1/p' | sort -V | tail -n 1)
+EXAMPLE_NO_SYNC_SIMULATOR ?= iPhone 17e,$(LATEST_IOS_SIMULATOR_OS)
+EXAMPLE_SYNC_SIMULATOR ?= iPhone 17 Pro Max,$(LATEST_IOS_SIMULATOR_OS)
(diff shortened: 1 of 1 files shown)


mohamede1945:
Jazak Allah khyrn

A-Levin:
Wa iyyak

PR #1043, opened 2026-09-10
Show non-default audio playback options in the banner

mohamede1945:
Show a red indicator on the audio banner’s More button and a localized summary whenever playback speed or verse/range repetition differs from the defaults. For example, `0.5× · verse ×3 · range ×2`; indefinite repetition uses `∞`.

The summary is anchored below the More icon, with symmetric padding that keeps controls vertically centered. Shared audio control views centralize the minimum tap area. Move `Runs` into `QuranAudio` and QueuePlayer into Domain so NoorUI uses the shared model without depending on the player.

Validation:
- Audio, advanced audio options, and NoorUI tests passed with sync enabled and disabled after the module move.
- All 43 NoorUI tests passed again after the final UI refactor.
- SwiftFormat lint passed.
- Rebuilt and visually checked previews on iOS 26.0 and 26.5, including default settings, repetition, and a right-to-left layout with larger text.

| iOS 26.0 | iOS 26.5 |
| --- | --- |
| Audio options preview on iOS 26.0 | Audio options preview on iOS 26.5 |


UI/NoorUI/Sources/Features/AudioBanner/AudioBannerViewUI.swift
```diff
@@ -55,20 +55,22 @@ public struct AudioBannerActions {
 public struct AudioBannerViewUI: View {
     private let state: AudioBannerState
     private let actions: AudioBannerActions
+    private let options: AudioOptionsSummary
     @ScaledMetric private var maxWidth = 500
 
-    public init(state: AudioBannerState, actions: AudioBannerActions) {
+    public init(state: AudioBannerState, actions: AudioBannerActions, options: AudioOptionsSummary = .init()) {
```
mohamede1945: delete the default options value here

UI/NoorUI/Sources/Features/AudioBanner/AudioBannerViewUI.swift
```diff
@@ -101,67 +103,27 @@ private struct AudioPlaying: View {
     let paused: Bool
     let currentRate: Float
     let actions: AudioBannerActions
+    let options: AudioOptionsSummary
 
     @Environment(\.layoutDirection) private var layoutDirection
     @ScaledMetric private var minimumRateSpacing = 4.0
+    @ScaledMetric private var summarySpacing = 4.0
```
mohamede1945: we shouldn't add summarySpacing here or update the padding, we should just have audioOptionsbutton add a little bit of extra height so everything works fine overall

UI/NoorUI/Sources/Features/AudioBanner/AudioOptionsSummary.swift
```diff
@@ -0,0 +1,40 @@
...
+        }
+        if rangeRuns != .finite(1) {
+            parts.append(String(format: l("audio.summary.range"), count(rangeRuns)))
+        }
+        return parts.joined(separator: " · ")
+    }
+
+    private func count(_ runs: Runs) -> String {
+        switch runs {
+        case .finite(let count): return "×" + count.formatted(.number.locale(Locale.current.fixedLocaleNumbers()))
```
mohamede1945: we have a formatter we can use, right? because it can handle arabic numbers I believe

UI/NoorUI/Sources/Features/AudioBanner/AudioBannerViewUI.swift
```diff
@@ -260,24 +221,25 @@ private struct AudioPlaying: View {
...
         ZStack {
             HStack {
                 Button(action: actions.play) {
-                    NoorSystemImage.play.image
-                        .padding()
+                    AudioControlLabel {
+                        NoorSystemImage.play.image
+                            .padding(.vertical, options.hasNonDefaultValues ? 0 : nil)
+                    }
+                    .padding(.horizontal)
```
mohamede1945: why do we have this code rather than the .padding() we had previously? do we need to depend on options.hasNondefaultValues? I feel this is the job of the more button

PR #1055, opened 2026-09-15
Show interactive ayah annotations in translation capsules

mohamede1945:
Stacked on #1054.

Translation ayah references now use a capsule with a divider and live note, collection, and colored reading bookmark icons. Verses without annotations keep the number-only capsule. Tapping any ayah-number capsule opens the existing ayah menu at that verse, with a minimum 44-point touch target.

Annotation state is filtered to displayed verses and deduplicated, and the icon rendering is shared with image-page annotation buttons. Localized numbers, theme colors, and Dynamic Type support are preserved.

Validation:
- Translation and Quran text tests pass, including live annotation updates, removal, displayed-verse changes, and suppression of unrelated updates.
- Sync and no-sync builds, SwiftFormat lint, and diff checks pass.
- Verified on the booted iPhone 17 Pro Max simulator: annotated and unannotated ayah-number capsules open the corresponding ayah menu.

| Annotation capsules | Ayah menu after tapping |
| --- | --- |
| Translation ayah capsules with annotations | Ayah menu opened from a reading bookmark capsule |


Features/QuranContentFeature/Sources/PagesView.swift
```diff
@@ -32,7 +32,13 @@ struct PagesView: View {
                         viewModel.deps.imageDataSourceBuilder.build(at: page)
                         #endif
                     case .translation:
+                        #if QURAN_SYNC
+                        viewModel.deps.translationDataSourceBuilder.build(at: page) {
+                            viewModel.onAnnotatedAyahTapped($0, at: $1)
```
mohamede1945: rename onAnnotatedAyahTapped to be onAyahNumberTapped and make it work whether there is annotation or not

